PR #77, opened 2014-03-28
Adding custom validators to completers

thijsdezoete:


kislyuk:
You had an API-breaking bug in your PR. Fixed in 105b1f1.

thijsdezoete:
Hi @kislyuk, thanks for the fix. And of course, sorry for breaking the api.

PR #93, opened 2015-02-07
Add REPL subparsers support

lisongmin:
This commit add subparsers support for REPL with readline module.
And multi level supparsers also supported.

tested pass under python2.7 and python3.4

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -227,11 +227,12 @@ class CompletionFinder(object):
         We save all active ArgumentParsers to extract all their possible option names later.
         '''
         self.active_parsers = []
-        visited_actions = []
+        self.visited_positionals = []
 
         completer = self
 
         def patch(parser):
+            completer.visited_positionals.append(parser)
             completer.active_parsers.append(parser)
 
             if isinstance(parser, IntrospectiveArgumentParser):
@@ -255,8 +256,6 @@ class CompletionFinder(object):
                         debug('\torig class:', self._orig_class)
                         debug('\torig callable:', self._orig_callable)
 
-                        visited_actions.append(self)
-
                         if not completer.completing:
                             self._orig_callable(parser, namespace, values, option_string=option_string)
                         elif self._orig_class == argparse._SubParsersAction:
@@ -264,6 +263,9 @@ class CompletionFinder(object):
                             patch(self._name_parser_map[values[0]])
                             self._orig_callable(parser, namespace, values, option_string=option_string)
                         elif self._orig_class in safe_actions:
+                            if not self.option_strings:
+                                completer.visited_positionals.append(self)
+
                             self._orig_callable(parser, namespace, values, option_string=option_string)
 
                 action._orig_class = action.__class__
@@ -273,7 +275,7 @@ class CompletionFinder(object):
         patch(self._parser)
 
         debug("Active parsers:", self.active_parsers)
-        debug("Visited actions:", visited_actions)
+        debug("Visited positionals:", self.visited_positionals)
 
         return self.active_parsers
 
@@ -285,9 +287,7 @@ class CompletionFinder(object):
         This method is exposed for overriding in subclasses; there is no need to use it directly.
         '''
         def get_subparser_completions(parser, cword_prefix):
-            return [subcmd for action in parser._actions
-                    if isinstance(action, argparse._SubParsersAction)
-                    for subcmd in action.choices.keys()
+            return [subcmd for subcmd in parser.choices.keys()
                     if subcmd.startswith(cword_prefix)]
 
         def get_option_completions(parser, cword_prefix):
@@ -296,13 +296,18 @@ class CompletionFinder(object):
                     for option in action.option_strings
                     if option.startswith(cword_prefix)]
 
-        def complete_active_option(parser, cword_prefix, parsed_args, completions):
+        def complete_active_option(parser, next_positional,
+                                   cword_prefix, parsed_args, completions):
             debug("Active actions (L={l}): {a}".format(l=len(parser.active_actions), a=parser.active_actions))
 
             # Only run completers if current word does not start with - (is not an optional)
             if len(cword_prefix) == 0 or cword_prefix[0] not in parser.prefix_chars:
                 for active_action in parser.active_actions:
                     if not active_action.option_strings: # action is a positional
+                        # show next positional only
+                        if active_action != next_positional:
+                            continue
+
                         if action_is_satisfied(active_action) and not action_is_open(active_action):
                             debug("Skipping", active_action)
                             continue
@@ -347,27 +352,49 @@ class CompletionFinder(object):
             return completions
 
         completions = []
-        """
-        for active_parser in active_parsers[:-1]:
-            debug('active_parser:', active_parser)
-            if self.always_complete_options or (len(cword_prefix) > 0 and cword_prefix[0] in active_parser.prefix_chars):
-                completions += get_option_completions(active_parser, cword_prefix)
-                debug('completions options:', completions)
-
-            completions = complete_active_option(active_parser, cword_prefix, parsed_args, completions)
-        """
 
         debug('all active parser:', active_parsers)
         active_parser = active_parsers[-1]
         debug('active_parser:', active_parser)
-
-        completions += get_subparser_completions(active_parser, cword_prefix)
         completions += get_option_completions(active_parser, cword_prefix)
-        debug('completions options:', completions)
-        completions = complete_active_option(active_parser, cword_prefix, parsed_args, completions)
-        debug('completions options:', completions)
+        debug('optional options:', completions)
+
+
+        next_positional = self._get_next_positional()
+        if isinstance(next_positional, argparse._SubParsersAction):
+            completions += get_subparser_completions(next_positional, cword_prefix)
+
+        debug('next_positional:', next_positional)
+        completions = complete_active_option(active_parser, next_positional,
+                                             cword_prefix, parsed_args,
+                                             completions)
+        debug('active options:', completions)
 
         return completions
+
+    def _get_next_positional(self):
+        """
+        Get the next positional action if exist.
+        """
+        active_parser = self.active_parsers[-1]
+        last_positional = self.visited_positionals[-1]
+
+        all_positional = active_parser._get_positional_actions()
+        if not all_positional:
+            return None
+
+        if active_parser == last_positional:
+            return all_positional[0]
+
+        i = 0
+        for i in range(len(all_positional)):
+            if all_positional[i] == last_positional:
+                break
+
+        if i + 1 < len(all_positional):
+            return all_positional[i + 1]
+
+        return None
 
     def filter_completions(self, completions):
         '''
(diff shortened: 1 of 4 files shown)


argcomplete/__init__.py
```diff
@@ -348,6 +343,30 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
...
+        completions = []
+        """
+        for active_parser in active_parsers[:-1]:
+            debug('active_parser:', active_parser)
+            if self.always_complete_options or (len(cword_prefix) > 0 and cword_prefix[0] in active_parser.prefix_chars):
+                completions += get_option_completions(active_parser, cword_prefix)
+                debug('completions options:', completions)
+
+            completions = complete_active_option(active_parser, cword_prefix, parsed_args, completions)
+        """
```
kislyuk: Why is this code block still here but commented?

I also worry that this seems like it might break completions in situations where more than one parser is in the active_parsers stack, and the last one is not the one that will consume the word under the cursor. I'll have to think more about when that's possible though (and write a test).

kislyuk:
Thanks, this is great. Please see the comment inline.

argcomplete/__init__.py
lisongmin: should we ignore words after cursor? cut it before split_line like this https://github.com/lisongmin/argcomplete/commit/f605e3a49ba993ac80aaca40bbe5f223457ccf13

lisongmin:
I walk through the code, words after cursor is cut in split_line. This should ensure last active parser is what we needed.

wapiflapi:
Tested this on my REPL usecase and didn't notice any regressions so far. Looks good.

lisongmin:
This pull can not work if positional with nargs > 1, sorry.

kislyuk:
I tested and positionals with nargs > 1 seem to work at the top level. Do you mean in subparsers?

argcomplete/__init__.py
```diff
@@ -284,30 +289,46 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
...
-                            subparser_activated = True
-                    if subparser_activated:
-                        # Parent parser completions are not valid in the subparser, so flush them
-                        completions = []
-                    else:
-                        completions += [subcmd for subcmd in action.choices.keys() if subcmd.startswith(cword_prefix)]
-                elif self.always_complete_options or (len(cword_prefix) > 0 and cword_prefix[0] in parser.prefix_chars):
-                    completions += [option for option in action.option_strings if option.startswith(cword_prefix)]
+        def get_subparser_completions(parser, cword_prefix):
+            """ """
```
kislyuk: Please remove empty docstrings.

argcomplete/__init__.py
```diff
@@ -284,30 +289,46 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
...
+            def filter_aliases(metavar, dest, prefix):
+                if not metavar:
+                    return dest
+
+                a = metavar.replace(',', '').split()
```
kislyuk: Can you explain what is happening here?

argcomplete/__init__.py
```diff
@@ -284,30 +289,46 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
...
+                return ' '.join(x for x in a if x.startswith(prefix))
+
+            self._display_completions.update(
+                [[filter_aliases(action.metavar, action.dest, cword_prefix),action.help]
+                 for action in parser._get_subactions()])
```
kislyuk: The construct that you use to add completions to _display_completions is really hard to read. Please just use a for loop and append to _display_completions. Also, if you're going to use tuples like this, please use NamedTuples so that the semantic meaning of the contents is apparent.

argcomplete/__init__.py
```diff
@@ -326,11 +347,21 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
                                 # This means the current action will fail to parse if the word under the cursor is not given
                                 # to it, so give it exclusive control over completions (flush previous completions)
                                 debug("Resetting completions because", active_action, "is unsatisfied")
+                                self._display_completions = {}
                                 completions = []
                         if callable(completer):
-                            completions += [c for c in completer(prefix=cword_prefix, action=active_action,
+                            comp = [c for c in completer(prefix=cword_prefix, action=active_action,
```
kislyuk: Please call this something more appropriate, maybe completions_from_callable.

argcomplete/__init__.py
```diff
@@ -345,11 +377,64 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
...
+                                self._display_completions.update([[x, ''] for x in comp])
+                                completions += comp
                         except subprocess.CalledProcessError:
                             pass
+
+            return completions
+
+        completions = []
+
+        debug('all active parser:', active_parsers)
```
kislyuk: s/parser/parsers/

argcomplete/__init__.py
```diff
@@ -345,11 +377,64 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
...
+        active_parser = active_parsers[-1]
+        debug('active_parser:', active_parser)
+        completions += get_option_completions(active_parser, cword_prefix)
+        debug('optional options:', completions)
+
+
```
kislyuk: Please do not use double line breaks.

argcomplete/__init__.py
```diff
@@ -345,11 +377,64 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
...
+        debug('active options:', completions)
+
+        self._display_completions.pop('', None)
+        debug('display completions:', self._display_completions)
+
         return completions
 
+    def _get_next_positional(self):
+        """
+        Get the next positional action if exist.
```
kislyuk: s/if exist/if it exists/

argcomplete/__init__.py
```diff
@@ -345,11 +377,64 @@ def collect_completions(self, active_parsers, parsed_args, cword_prefix, debug):
...
+        """
+        active_parser = self.active_parsers[-1]
+        last_positional = self.visited_positionals[-1]
+
+        all_positional = active_parser._get_positional_actions()
```
kislyuk: I think this is better called "all_positionals"?

argcomplete/__init__.py
```diff
@@ -447,6 +532,86 @@ def complete(self, text, state):
...
+            readline.set_completer_delims(' \t\n=')
+            readline.parse_and_bind('"\M-OA": previous-history')
+            readline.parse_and_bind("tab: complete")
+            result = input("prompt> ")
+
+        (Use ``raw_input`` instead of ``input`` on Python 2, or use `eight <https://github.com/kislyuk/eight>`_).
+        """
+        debug('text:[{0}]'.format(text), ', state:', state)
+        if state == 0:
+            # `text` may empty, so we read line from readline.
```
kislyuk: s/may/may be/

argcomplete/__init__.py
```diff
@@ -447,6 +532,86 @@ def complete(self, text, state):
...
+            debug('matches', self.matches)
+
+        if state < len(self.matches):
+            return self.matches[state]
+        else:
+            return None
+
+    def get_display_completions(self):
+        """
+        This function return complete key and description for customize display hook.
```
kislyuk: s/return/returns/

kislyuk: Please update this docstring with a more specific description of the returned contents. e.g. "This function returns a mapping of option names to their help strings for displaying to the user."

lisongmin:
nargs='+'  not work with my expected, 
it can not show two positionals's choices together if  the first one is setting already.

kislyuk:
Can you explain with an example? It's not possible to have two positionals with nargs=+ active at the same position, because the parser wouldn't know which one to parse for.

lisongmin:
I means one nargs='+', it's my mistake and fixed at last commit.
Please ses the comment inline.

lisongmin:
below test can not pass:
```python
    def test_nargs_one_or_more(self):
        def make_parser():
            parser.add_argument('h1', choices=['c', 'd'])
            parser.add_argument('var', choices=['bus', 'car'], nargs='+')
            parser.add_argument('value', choices=['orange', 'apple'])
            return parser

        expected_outputs = (("prog ", ['c', 'd', '-h', '--help']),
            ("prog c bu", ['bus ']),
            ("prog c bus ", ['bus', 'car', 'apple', 'orange', '-h', '--help']),
            ("prog c bus car ", ['bus', 'car', 'apple', 'orange', '-h', '--help']),
            ("prog c bus appl", ['apple ']),
            ("prog c bus apple ", ['-h', '--help']),
            )

        for cmd, output in expected_outputs:

```
After input 'var' positional more than one, it can not show 'var' choices.
Active actions is 'value' positional now:
```
Active actions (L=1): [IntrospectAction(option_strings=[], dest='value', nargs=None, const=None, default=None, type=None, choices=['orange', 'apple'], help=None, metavar=None)]
```

kislyuk:
I see, thanks. This is great. Could you please add this test to the suite, but mark it with an expectedFailure decorator (https://docs.python.org/2/library/unittest.html#unittest.expectedFailure)?

lisongmin:
testcases are commited. 
I add another failure testcase that positional can not complete again in parser reuse.
It seems "num_consumed_args" can not reset, so we can not get the correct value.

lisongmin:
I revert the testcase for it cause failure on pypy3. I don't known why it cause failure.
Test in my local computer is all right.

```
The command "make test" exited with 0.

6.02s$ coverage run --omit='*/site-packages/*,*/distutils/*' ./test/test.py

.........x..E..x.

======================================================================

ERROR: test_subparsers (__main__.TestArgcomplete)

----------------------------------------------------------------------

File "./test/test.py", line 81, in run_completer

exit_method=sys.exit, **kwargs)

File "/opt/python/pypy3-2.4.0/lib-python/3/unittest/case.py", line 557, in assertRaises

callableObj(*args, **kwargs)

File "/home/travis/build/kislyuk/argcomplete/argcomplete/__init__.py", line 197, in __call__

output_stream.flush()

IOError: [Errno 9] Bad file descriptor

During handling of the above exception, another exception occurred:

File "./test/test.py", line 207, in test_subparsers

File "./test/test.py", line 83, in run_completer

return t.read().decode(locale.getpreferredencoding()).split(IFS)

IOError: [Errno 9] Bad file descriptor
```

kislyuk:
I believe that's a transient failure. Please put them back, those failures are not the fault of those test cases.

lisongmin:
OK

kislyuk:
OK, thanks, I see what's going on now. The tests pass, but the coverage collector is not liking the expected test failures. I'll replace them with the `@skip` decorator after merging.

kislyuk:
Thanks. I still have problems with parts of this PR (mostly semantics of variable naming, etc.), but I'm merging it now and will work through them myself in the main tree. In the future, I will need any PR like this to be broken down into smaller, easier to understand chunks.

kislyuk:
In retrospect, I should not have merged this. The code in this PR has severe compositional and logic deficiencies which led to multiple defects down the line.

PR #95, opened 2015-02-16
Add note about metavar flag

asmeurer:
This tripped me up. I had a choices completer with a large number of
completions and it was polluting the argparse help output.


kislyuk:
I'm not saying no, but this is purely an argparse behavior, nothing to do with argcomplete, right?

asmeurer:
Yes it is. It's purely a documentation issue. The docs suggest using `choices`, which works great, except for this little issue, which I didn't even expect to happen, and when I noticed it, I had to research how to fix it. It would have saved me some time if this were in the docs when I read them.

kislyuk:
Sounds good, thanks.

PR #98, opened 2015-03-01
Added completer for directories

drmalex07:
Hello, and congrats for the nice work. Lately, I have been using `argcomplete` more and more for my administrative Python scripts.

I sometimes came across the situation where i needed directory-only completion (e.g. like `complete -d`). I have created a completer class to address this, and i thought that it could be usefull to someone.

I did'nt use the subprocess approch that you have followed in `FilesCompleter`, because the task is quite simple and can be easily performed with plain `os.path` tools.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -183,6 +183,30 @@ class TestArgcomplete(unittest.TestCase):
                 fp.write('test')
             self.assertEqual(set(fc('a')), set(['abcdef/', 'abcaha/', 'abcxyz']))
 
+    def test_directory_completion(self):
+        from argcomplete.completers import DirectoriesCompleter
+        completer = DirectoriesCompleter()
+        c = lambda prefix: set(completer(prefix))
+        with TempDir(prefix='test_dir', dir='.') as temp_dir:
+            # Create some temporary dirs and files (files must be ignored)
+            os.makedirs(os.path.join('abc', 'baz'))
+            os.makedirs(os.path.join('abb', 'baz'))
+            os.makedirs(os.path.join('abc', 'faz'))
+            os.makedirs(os.path.join('def', 'baz'))
+            with open('abc1', 'w') as fp1, open('def1', 'w') as fp2:
+                fp1.write('A test')
+                fp2.write('Another test')
+            # Test completions
+            self.assertEqual(c('a'), set(['abb/', 'abc/']))
+            self.assertEqual(c('ab'), set(['abc/', 'abb/']))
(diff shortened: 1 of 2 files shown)


kislyuk:
Looks good overall, please remove trailing semicolons from your lines.

kislyuk:
Also, please add a test for your completer. You should be able to just copy/paste one of the existing tests and build a custom parser that uses this completer. To make it real, you might want to use https://docs.python.org/3.4/library/tempfile.html functions.

drmalex07:
Hello @kislyuk. I 've removed the unneeded semicolons.

Regarding to the tests, i suppose we need a test similar to the one at
https://github.com/kislyuk/argcomplete/blob/master/test/test.py#L174 ??

kislyuk:
Correct. Thanks.

drmalex07:
Hello @kislyuk. I've added a relevant test. Please, check if this is adequate.

And one small question: Before adding this test, Travis failed at the phase of PyPy3 (while checking coverage). Has this failure anything to do with this, i mean are these tests strictly required for the coverage checks to pass ?

kislyuk:
Thanks. I'll check this later today. Don't worry about the failure during
coverage checking - that is a bug in the test suite that got recently
introduced, and is not your fault. Sorry about that.

-ak

drmalex07:
Hello @kislyuk , is there anything (from my side) pending on this issue ?

test/test.py
```diff
@@ -183,6 +183,30 @@ def test_file_completion(self):
...
+            # Test completions
+            self.assertEqual(c('a'), set(['abb/', 'abc/']))
+            self.assertEqual(c('ab'), set(['abc/', 'abb/']))
+            self.assertEqual(c('abc'), set(['abc/']))
+            self.assertEqual(c('abc/'), set(['abc/baz/', 'abc/faz/']))
+            self.assertEqual(c('d'), set(['def/']))
+            self.assertEqual(c('def/'), set(['def/baz/']))
+            self.assertEqual(c('e'), set([]))
+            self.assertEqual(c('def/k'), set([]))
+        return
```
kislyuk: The `return` is unnecessary, please remove it.

kislyuk:
Sorry about the delay, thanks for your patience @drmalex07. I went ahead and merged this. I'll try to roll a release soon.

drmalex07:
No problem at all, i only asked if anything was missing. 

I am pretty sure we all have pressing deadlines on our heads!

PR #105, opened 2015-04-17
Add specifiable complete arguments at global activation/registration

Louis-Kenzo:
Makes it possible to remove defaults using global activation and no just with registration. Also enables generically customizing complete arguments. Feedback is given when non-default arguments are specified when globally activating.

The `--no-defaults` option is kept for backward compatibility and ease of use.

Changes are best seen running `activate-global-python-argcomplete --dest=-` with various options and looking at the last line in the ouput, the call to complete. e.g:
* `activate-global-python-argcomplete --dest=-`<br/>
  ↳ `complete -o nospace -o default -o bashdefault -D -F _python_argcomplete_global`
* `activate-global-python-argcomplete --dest=- --no-defaults`<br/>
  ↳ `complete -o nospace -o bashdefault -D -F _python_argcomplete_global`
* `activate-global-python-argcomplete --dest=- --complete-arguments --o nospace`<br/>
  ↳ `complete -o nospace -D -F _python_argcomplete_global`
* `activate-global-python-argcomplete --dest=- --complete-arguments`<br/>
  ↳ `complete -D -F _python_argcomplete_global`

In summary, the call to complete in `python-argcomplete.sh` is removed, and appended to the file with appropriate arguments after the file is copied. The same options are added to `register-python-argcomplete` for consistency.

Note: the difference remains that by default the register script passes neither `-D` nor `-o bashdefault` to `complete`, whereas the activate script does pass them.

### DIFF
--- a/scripts/activate-global-python-argcomplete
+++ b/scripts/activate-global-python-argcomplete
@@ -8,13 +8,17 @@
 Activate the generic bash-completion script for the argcomplete module.
 '''
 
-import os, sys, argparse, argcomplete, shutil
+import os, sys, argparse, argcomplete, shutil, fileinput
 
 parser = argparse.ArgumentParser(description=__doc__,
                                  formatter_class=argparse.RawDescriptionHelpFormatter)
 
 dest_opt = parser.add_argument("--dest", help="Specify the bash completion modules directory to install into", default="/etc/bash_completion.d")
 parser.add_argument("--user", help="Install into user directory (~/.bash_completion.d/)", action='store_true')
+parser.add_argument("--no-defaults", dest="use_defaults", action="store_false", default=True,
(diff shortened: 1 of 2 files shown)


kislyuk:
Thank you for your contribution. This makes sense, but I prefer that the file `argcomplete/bash_completion.d/python-argcomplete.sh` remain ready to be copied or sourced manually, so if you don't mind, please change the PR to add the original `complete` call back in, and to have `scripts/activate-global-python-argcomplete` replace the line instead of adding it.

Louis-Kenzo:
The commit was rebased to copy `argcomplete/bash_completion.d/python-argcomplete.sh` and replace the call to `complete` afterwards as you suggested.

kislyuk:
Didn't know about fileinput. Cool!

PR #107, opened 2015-04-25
fix #106: doesn't complete choices with type other than str.

### LINKED ISSUE #106: Don't complete when add_argument has both type and choices set to other class than str.
I've notice that argcomplete will complete `choices` if each choice has type of `str`, so this code can be completed:

```py
parser.add_argument('x', type=int, choices=['4','8','15','16','23','42'])      
print(vars(parser.parse_args()))
```

But failed with actual calling, because `parse_args()` do parse input string into the specify `type` first, then check if an argument met constrain with `choices`, for instance:

```
$ ./it.py 42
usage: it.py [-h] {4,8,15,16,23,42}
it.py: error: argument x: invalid choice: 42 (choose from '4', '8', '15', '16', '23', '42')
```

While this corrected code, that has constrain of `choices` according to its `type`, is running fine.

```py
parser.add_argument('x', type=int, choices=[4,8,15,16,23,42])
print(vars(parser.parse_args()))
```

But failed to complete.

neizod:


kislyuk:
Thanks for the report. This looks like it's generally the wrong thing to do for non-ascii strings. I'll play around with this and see if the test suite needs to be extended to reflect that.

PR #110, opened 2015-05-28
Made context managers exception-safe.

msiedlarek:
Context managers used to suppres stdout and stderr output were not
exception safe. When an exception occurred in wrapped code (such as
`SystemExit` from `parse_args()`) the correct file objects were not put
back. This has caused stderr to be permanently muted in several use
cases, specifically in realine-based programs.

This patch fixes the issue by following recommendations from
`contextlib` documentation, see:
https://docs.python.org/2/library/contextlib.html#contextlib.contextmanager


kislyuk:
Yep, my fault. Thanks for the fix!

PR #118, opened 2015-10-27
Allow overridden _SubParsersAction

kashizui:
Use an `issubclass` check rather than `==` to allow subclassing of `argparse._SubParsersAction` in applications.

Allows useful extensions such as aliasing of subcommands:
https://gist.github.com/sampsyo/471779


kislyuk:

PR #120, opened 2015-11-15
Consider possible subclass of argparse._SubParsersAction

jeromelebleu:
This pull request allows to auto-complete actions which subclass `argparse._SubParsersAction`.

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -284,7 +284,7 @@ class CompletionFinder(object):
 
                         if not completer.completing:
                             self._orig_callable(parser, namespace, values, option_string=option_string)
-                        elif self._orig_class == argparse._SubParsersAction:
+                        elif issubclass(self._orig_class, argparse._SubParsersAction):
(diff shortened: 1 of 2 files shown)


kislyuk:
Can you explain how this relates to https://github.com/kislyuk/argcomplete/pull/118?

jeromelebleu:
Sorry again, I tried with the package in Debian Jessie which is version 0.8.1 and so doesn't come with that fix.
I test with #118 only and I confirm that it is working and enough, my other modifications are not useful. Thanks!

kislyuk:
OK, thanks for the clarification.

kashizui:
Hi @kislyuk , sorry I missed this discussion—it looks like #118 has not actually been merged yet, has it been patched onto a branch other than `master`?
Thanks!

kislyuk:
You are right, it wasn't. I just merged it. I'm not sure why I thought it was. Please let me know if there's anything else to be done with the changes here (ideally I would also put in a test, but I can't at the moment).

kislyuk:

PR #124, opened 2016-02-13
Support parsed_args in custom completers with missing args

dankilman:
Currently, if a custom completer makes use of `parsed_args` but not enough arguments have been supplied (either positional or `required` optional arguments), the default implementation raises an error. This logic is derived from the original argparse implementation (and copied in `my_argparse.py`).
This PR comments out the lines making this validation, as the validation is not required for argcomplete to populate the namespace for arguments that has been supplied thus far.
The test fails for obvious reason if the lines are not commented out, and passes if they are.
While this solution may seem naive, it seems to be correct.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,7 @@ nosetests.xml
 
 # vim python-mode plugin
 .ropeproject
+
+# IntelliJ IDEA / PyCharm project files
+.idea
+*.iml
--- a/argcomplete/my_argparse.py
+++ b/argcomplete/my_argparse.py
@@ -28,6 +28,7 @@ def action_is_open(action):
 class IntrospectiveArgumentParser(ArgumentParser):
     ''' The following is a verbatim copy of ArgumentParser._parse_known_args (Python 2.7.3),
     except for the lines that contain the string "Added by argcomplete".
+    Lines that contain the string "Disabled by argcomplete" have been commented out.
     '''
 
     def _parse_known_args(self, arg_strings, namespace):
@@ -206,7 +207,7 @@ class IntrospectiveArgumentParser(ArgumentParser):
                     self.active_actions.append(action)  # Added by argcomplete
                 args = arg_strings[start_index: start_index + arg_count]
                 start_index += arg_count
-                action.num_consumed_args = len(args)
+                action.num_consumed_args = len(args)   # Added by argcomplete
                 take_action(action, args)
 
             # slice off the Positionals that we just parsed and return the
@@ -261,14 +262,16 @@ class IntrospectiveArgumentParser(ArgumentParser):
 
         if positionals:
             self.active_actions.append(positionals[0])  # Added by argcomplete
-            self.error(_('too few arguments'))
+            # self.error(_('too few arguments'))  # Disabled by argcomplete
 
-        # make sure all required actions were present
-        for action in self._actions:
-            if action.required:
-                if action not in seen_actions:
-                    name = _get_action_name(action)
-                    self.error(_('argument %s is required') % name)
+        # Begin disabled by arg complete
+        # # make sure all required actions were present
+        # for action in self._actions:
+        #     if action.required:
+        #         if action not in seen_actions:
+        #             name = _get_action_name(action)
+        #             self.error(_('argument %s is required') % name)
+        # End disabled by argcomplete
 
         # make sure all required groups had one option present
         for group in self._mutually_exclusive_groups:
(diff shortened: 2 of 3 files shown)


kislyuk:
Thanks for reporting. The concept is solid, but my fear is that this change enables completion to continue running in situations where the resulting string will raise an error in actual parsing, because not all positionals have been satisfied.

In other words, argcomplete's first order of business is to try as hard as possible to stick to argparse's default behavior, including indicating that parsing will fail by refusing to supply any more completions when mashing `<tab>`.

If you can think of a way to show that behavior is kept intact - for example, only skip the assertion if currently completing a positional (??? I need to think about this more) - then I'll be fine with this change.

(Cosmetic nitpick: "disabled" should probably be used instead of "removed")

dankilman:
Thanks for the quick reply. 
To my understanding, `ArgumentParser` is only patched during completion. This combined with the fact that you can currently specify completers for "partial" argument list, makes your comment no so clear to me. For example, a slightly modified version of the test I added
``` python
def test_subparser_completers(self):
    def c_depends_on_positional_arg1(prefix, parsed_args, **kwargs):
        return ['SOME_VALUE']

    def c_depends_on_optional_arg5(prefix, parsed_args, **kwargs):
        return ['SOME_OTHER_VALUE']

    def make_parser():
        parser = ArgumentParser()
        subparsers = parser.add_subparsers()
        subparser = subparsers.add_parser('subcommand')
        subparser.add_argument('arg1')
        subparser.add_argument('arg2').completer = c_depends_on_positional_arg1
        subparser.add_argument('arg3').completer = c_depends_on_optional_arg5
        subparser.add_argument('--arg4', required=True)
        subparser.add_argument('--arg5')
        return parser

    expected_outputs = (
        ("prog subcommand val1 ", ["SOME_VALUE", "--arg4", "--arg5", "-h", "--help"]),
        ("prog subcommand val1 val2 --arg5 val5 ", ["SOME_OTHER_VALUE", "--arg4", "--arg5", "-h", "--help"]),
    )

    for cmd, output in expected_outputs:
```
Is currently supported by `argcomplete`, even though it allowing this completion will still result in a command that `argparse` will refuse to accept.
The only thing that is not supported, is populating `parsed_args` in these cases.
I'll appreciate it if you elaborated. Thanks.
(I'll modify the wording to use Disabled)

dankilman:
Regarding the assertion skipping suggested, for the first line commented out, this is currently the case with this PR, i.e. `positionals` will only be non empty if there are positionals left. 
Regarding the required optional arguments disabling (the chuck of lines commented out), a similar rational applies, i.e. a user may currently wish to complete an optional argument based on some previous argument and has yet to fill the required optional argument. argcomplete supports this use case, it simply does not populate `parsed_args` as mentioned previously.

kislyuk:
OK, after thinking about it longer with more coffee ;) I understand now what's going on.

The general MO of the patched argparse is to operate argparse as normal, but install the minimal sufficient instrumentation needed to collect information about the currently active action(s) (those responsible for parsing the argument under the cursor), and - secondarily - the results of parsing so far.

The intent is to collect this information out-of-band, so that even if argparse crashes (which is expected in the presence of incomplete arguments), the information becomes available.

What you discovered is that we can't rely on argparse to collect the parsing results (`namespace`) out-of-band when running subparsers, so if an argparse crash occurs, they are not recovered and do not become available to the completer as promised by the API. This is the culprit snippet in argparse:
```
	# In case this subparser defines new defaults, we parse them
        # in a new namespace object and then update the original
        # namespace for the relevant parts.
        subnamespace, arg_strings = parser.parse_known_args(arg_strings, None)
        for key, value in vars(subnamespace).items():
            setattr(namespace, key, value)
```

I'm considering whether your solution is preferable to patching `argparse._SubParsersAction` to preserve the parsing results out-of-band.

dankilman:
Interesting. So a patched version may do something like this
``` python
subnamespace = Namespace()
try:
    subnamespace, arg_strings = parser.parse_known_args(arg_strings, subnamespace)
finally:
    for key, value in vars(subnamespace).items():
        setattr(namespace, key, value)
```
You are most certainly more familiar with the consequences of each choice. Both seem valid to me, albeit, the latter a little less hacky.

dankilman:
Are you still considering this PR?
I'd be happy to test out your proposed solution, although I'll need a small amount of guidance to follow what you had in mind.

kislyuk:
Yeah, for now let's just go with your solution.

dankilman:
Thanks!

kislyuk:

dankilman:
Just did, works great!

PR #130, opened 2016-05-19
Add ability to change default completer

avylove:
The default completer is hardcoded to FilesCompleter, so redefining it would require redefining FilesCompleter after importing argcomplete. This provides a cleaner way to replace the default completer.


kislyuk:
Good idea, thanks.

PR #132, opened 2016-06-04
Fix handling for nargs=REMAINDER

evanunderscore:
This adds handling for `nargs=argparse.REMAINDER` for both positionals and optionals.

Some things to note:
* The reset check has moved outside the loop (I'm assuming this could previously only be triggered once)
* Completers are no longer unconditionally skipped when the current argument looks like an optional

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -7,7 +7,7 @@ import os, sys, argparse, contextlib
 from . import completers, my_shlex as shlex
 from .compat import USING_PYTHON2, str, sys_encoding, ensure_str, ensure_bytes
 from .completers import FilesCompleter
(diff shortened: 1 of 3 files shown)


kislyuk:
This is great work. The tests are good enough proof of correctness, merged.

Your relocation of the reset clause is correct for greedy actions (per your nomenclature, where greedy means "must consume the next action").

More generally, there is another related use case that currently doesn't work. It involves multiple active actions (per my nomenclature, where active means "may consume the next action"). One example is in the test suite under `test_nargs_one_or_more` (currently marked as an expected failure). If you're interested, take a look and see if you can figure out how to fix it.

evanunderscore:
Thanks! I did notice that test case earlier and have been thinking about what could be done about that. I'll let you know if I get anywhere with it.

PR #133, opened 2016-06-05
Attempt to handle variable length positional nargs

evanunderscore:
This is an imperfect solution that chooses to show too many completions rather than too few. Note that this somewhat blurs the meaning of `active_actions` since we have to store the whole sequence of positionals in case we need to run completers for all of them.

The problem is that you can't know in advance which arguments belong to which action. For example, one of the original tests was `"prog c bus apple " => ["-h", "--help"]`. The user intends for `apple` to fill in `value` and hence no more positional arguments can be added. However, if the user were to then type `orange`, argparse would assign both `bus` and `apple` to `var` (in this case that happens to be invalid, but it won't always be). The solution I've taken is to always complete all positionals after `var` since anything typed past that point could eventually end up as one of those arguments.

I'm not 100% sure this is the best solution, but I've put up this PR for discussion.

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -343,12 +343,21 @@ class CompletionFinder(object):
             # Only run completers if current word does not start with - (is not an optional)
             return completions
 
+        complete_remaining_positionals = False
         # Use the single greedy action (if there is one) or all active actions.
         for active_action in greedy_actions or parser.active_actions:
             if not active_action.option_strings:  # action is a positional
-                if action_is_satisfied(active_action) and not action_is_open(active_action):
-                    debug("Skipping", active_action)
-                    continue
+                if action_is_open(active_action):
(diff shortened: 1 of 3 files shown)


kislyuk:
I think this is very close to perfect. I have one improvement in mind - don't add an action to active actions if it's preceded by an unsatisfied action with nargs > 0 - but I can play around with this and add it myself.

Thanks so much for your work on this!

kislyuk:
After thinking about this for a bit, I agree with you - it's intractable to try to guess which trailing actions to inactivate if there is an unbounded-nargs action active. So I think we're set here - great work!

evanunderscore:
Thanks! I think it may be possible to improve the handling of `nargs='?'` since it can only take 0 or 1 arguments. In this case, it's probably only necessary to to complete 1 extra argument when the action is open rather than all of them.

Issue #139, opened 2016-07-13
Clarify that filtering by prefix is not required in completers


kislyuk:
See #137

haifengkao:
really?

`argcomplete.autocomplete(parser, validator=lambda x, y: True)`
will not show anything.

I have to use 
`argcomplete.autocomplete(parser, validator=None)`
for argcomplete to work

haifengkao:
The problem seems to be related to zsh. 
        `output_stream.write('/usr/hello'.encode(sys_encoding))`
only works if the prefix of '/usr/hello' matches the suffix in the terminal

evanunderscore:
I feel like the choices completer (and probably the others too) should not filter their results and defer to the validator to do the filtering. As it stands, if I want to do something like case-insensitive completion, I have to override the choices completer because the results are already filtered before they get to the validator. Or am I misunderstanding the purpose of the validator?

kislyuk:
@haifengkao, it would be really awesome if a test could be added for this behavior. I'd like to create a pexpect test that would verify completions end-to-end in bash and zsh. Do you want to help?

@evanunderscore, the choices completer is presented as an example. The validator was added to this architecture after I first designed it, so I'm treating it as a general purpose filtering step. One of its functions is to filter the completions by prefix. The completer may also do so, for example if filtering will reduce the amount of work done (like when requesting completions from a remote API). Case insensitive completion is not generally supported. If you want it supported, please open another issue. We'll need to update the docs, think about how to communicate case insensitivity to completers, and add a CaseInsensitiveValidator.

haifengkao:
@kislyuk sure. what do you need?

kislyuk:
A test for the zsh behavior that you described (positive and negative test assertions). You may have to modify travis.yml to get zsh and pexpect installed.

haifengkao:
The current tests are pure python and they don't involve the shell at all. 
Do you want me to set up a new test framework for you?

kislyuk:
We would need a new test case that uses pexpect. I would like to have it added in a new class in `test/test.py`.

haifengkao:
I was doing a little research (http://www.zsh.org/mla/users/2013/msg00612.html).
Zsh has its own filtering system, which can be turned off by `compadd -U`. However, the link above suggests this `-U` option is not available in bash compatibility mode.

evanunderscore:
Assuming zsh is sufficiently covered by #10 and #141.

Issue #135, opened 2016-06-21
Can't run argcomplete on zsh


eyalev:
Hi,

I'm trying to run `argcomplete` with `zsh` on a fresh ubuntu:14.04 Docker container.

`# First trying using bash`
Enabled bash completion
Installed python, pip
pip install argcomplete
Created [describe_github_user.py](https://github.com/kislyuk/argcomplete/blob/master/docs/examples/describe_github_user.py) in `/root`
chmod +x describe_github_user.py
`activate-global-python-argcomplete`
Running `./describe_github_user.py --organization heroku --member <TAB>` works :+1:

`# Now trying zsh`
Installed `zsh`
Moved to `zsh` shell
In `.zshrc`:
````
autoload -Uz compinit bashcompinit
compinit
bashcompinit

source /etc/bash_completion.d/python-argcomplete.sh

eval "$(register-python-argcomplete /root/describe_github_user.py)"
````

Running `./describe_github_user.py --organization heroku --member <TAB>`

Not working :(

What I'm doing wrong?

Thanks

kislyuk:
I'm not a zsh expert, but I just tried the exact commands you pasted in my zsh and it worked fine. My zsh version is `zsh 5.2 (x86_64-apple-darwin15.4.0)`.

eyalev:
Ok, for some reason the `./` in the command was giving me problems.

I added the `root` directory to `PATH`:
`export PATH=/root:$PATH`

And then:
`describe_github_user.py --organization heroku --member <TAB>` worked. 

Yay :)

Maybe it's an Ubuntu specific thing..

If that's the case it might be worth mentioning in the README in some way. Could have saved me a lot of time.

----

Some backstory info: 

I knew it had to work somehow because I use Google's `gcloud` tool which uses `argcomplete`.

https://cloud.google.com/sdk/docs/quickstart-linux
https://github.com/eyalev/gcloud/blob/master/completion.zsh.inc

So I tried to find out what is different when I'm using `gcloud` and I got to the above "fix".

Issue #129, opened 2016-05-10
Argcomplete always sorts completions


julienbaley:
Even when one writes a custom completer, argcomplete will sort that list alphabetically before printing it.

Is this something that can be changed? Or is it not linked to argcomplete?

kislyuk:
Sorting is done by bash. I think there is nothing you can do to change it. See http://unix.stackexchange.com/questions/215937/disable-sorting-of-compreply-in-bash-complete-function

haifengkao:
zsh can do the custom sorting
http://stackoverflow.com/questions/15140396/zsh-completion-order

kislyuk:
Cool @haifengkao, if you have specific examples or a zsh-specific harness you want to add, let me know.

Issue #50, opened 2013-08-18
zsh escapes space after singleton completions


redstreet:
On unique completion possibilities, a trailing space is added. This has the nice effect of adding a space on the command line, but is not an ideal way to do so. zsh, for instance, is smart enough to notice that trailing space, and keeps escaping it automatically (so instead of '--help', zsh completes it as '--help\ ' which is incorrect).

IMHO, this is better handled by removing the '-o nospace' in:

```
complete -o nospace -o default -F _python_argcomplete "t"
```

...and removing these lines:
#### in **init**.py:

```
 If there's only one completion, and it's not open-quoted and doesn't end with a continuation char, add a space
 continuation_chars = '=/:'
if len(completions) == 1 and completions[0][-1] not in continuation_chars:
    if cword_prequote == '' and not completions[0].endswith(' '):
        completions[0] += ' '
```

Is there any reason not to do so that I'm not seeing?

evanunderscore:
Does it make sense to have "singleton completions that may be partial"? In the case of a directory name followed by a slash, should you not also return the contents of that directory? For example, if you return `dir/` and expect the user could use something in that directory, why not also return `dir/subdir/`, `dir/file`, etc?

kislyuk:
Yes, it does. Try completing a path of several directories in bash. Even if there is a unique completion, each `<TAB>` will yield just one directory completion at a time. That's an intentional UX design, and I'd like to conform to it.

kislyuk:
More generally, we can't reasonably return all completions in a tree hierarchy.

evanunderscore:
What I had assumed was happening is that when a directory would be returned as the only match, it also returned the contents of that directory, eg `dir/` and `dir/subdir` (but not `dir/subdir/file` or any deeper in the hierarchy), which would cause `dir/` to be filled in as the common prefix. However you are correct - even when `show-all-if-ambiguous` is turned on, bash will not display the contents of a directory until tab is pushed a second time.

PR #141, opened 2016-08-01
add native zsh support

haifengkao:
The codes of `_python_argcomplete_zsh` are mostly copied from `_bash_complete` in `bashcompinit`.
I don't really know how to do bash programming, so I made the pull request with copy-n-paste-programming and programming-by-coincidence techniques. I believe it will require a serious review.

It does
- use the native zsh completion system, so `autoload bashcompinit` is not required anymore (https://github.com/kislyuk/argcomplete/issues/135)
- preserve the completion order from the python scripts (zsh only https://github.com/kislyuk/argcomplete/issues/129)
- allow arbitrary completion even the prefix doesn't match the user input (zsh only https://github.com/kislyuk/argcomplete/issues/139)

### DIFF
--- a/scripts/register-python-argcomplete
+++ b/scripts/register-python-argcomplete
@@ -18,19 +18,101 @@ import sys
 import argparse
 
 shellcode = '''
-_python_argcomplete() {
-    local IFS='\013'
+_python_argcomplete_lst() {
+
     COMPREPLY=( $(IFS="$IFS" \
                   COMP_LINE="$COMP_LINE" \
                   COMP_POINT="$COMP_POINT" \
                   _ARGCOMPLETE_COMP_WORDBREAKS="$COMP_WORDBREAKS" \
                   _ARGCOMPLETE=1 \
                   "$1" 8>&1 9>&2 1>/dev/null 2>/dev/null) )
+
+    # bash use echo to return results
+    echo $COMPREPLY
+}
+
+_python_argcomplete_zsh() {
+  # need different IFS for bash and zsh. why?
+  local IFS='\n'
+
+  # use tab to cycle throught all possible matches
+  compstate[insert]=menu # no expand
+
+  local ret=1
+  local -a suf matches
+  local -x COMP_POINT COMP_CWORD
+  local -a COMP_WORDS COMPREPLY BASH_VERSINFO
+  local -x COMP_LINE="$words"
+  local -A savejobstates savejobtexts
+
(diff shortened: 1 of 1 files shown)


kislyuk:
As we discussed previously, we will need an end-to-end test harness in the test suite which can use pexpect to call the completion hook from bash and zsh, and test the expected result. We will also need the tests to pass. That is the starting point where I can begin to actually review this PR.

haifengkao:
Got it. The bash/zsh testing environment is a huge work.

haifengkao:
The strange space issue is explained in https://github.com/kislyuk/argcomplete/issues/50

evanunderscore:
#153 should give you a decent start on the pexpect tests. Running those tests now shows that these changes break bash.

haifengkao:
I may need some time to work on this one.

evanunderscore:
I had a quick look at getting the testing to work with zsh, and it's a bit more difficult. There's no zsh equivalent of `pexpect.replwrap.bash`, so you'll have to look at how that's implemented and build your own. I found that even after setting `PS1` it was still printing a bunch of other characters. However once you have that working you should be able to get all the bash tests free via subclassing.

sergei-dyshel:
What's the status of this PR? I see that it breaks bash support but is it ready to use for zsh-only completion? I've tried using bash completions scripts with `bashcompinit` but most of these completions just don't work.

haifengkao:
I don't really use bash, so I don't know what is supposed to work.
A simple `ls` completion works on bash.

kislyuk:
@sergei-dyshel, as you can see from the discussion above, the author of this PR has not been able to provide a test suite for the changes. The author also doesn't have a complete understanding of what's going on in the zsh completion loop, as is evident from the comments and complexity in the zsh completion harness that he wrote. Since I don't use zsh and don't have the time to learn it for this purpose, I can't contribute and this PR is stuck.

As far as I understand, zsh completion should work in argcomplete through the zsh bashcompinit layer (although we don't have tests for this in our test suite). Can you describe in detail the test that you performed that made you think it doesn't work?

haifengkao:
@kislyuk Thank you for your harsh but true comment. 
I think it is better to wait for someone who has the deep understanding of zsh completion to merge this PR. 
I use this PR on a daily basis. It will work, but I don't know why.

kislyuk:
zsh support was added in v3.0.0 using different shellcode.

PR #145, opened 2016-10-10
Ignore excluded arguments

### LINKED ISSUE #144: Exclusive groups are repeated on the compilation
When adding exclusive groups, the autocompletion will continue to offer all the members of the group although only one of them can be chosen.
Example:
```
#!/usr/bin/env python
import argparse
import argcomplete

group = parser.add_mutually_exclusive_group()
group.add_argument('--test', action='store_true')
group.add_argument('--no_test', action='store_true')
argcomplete.autocomplete(parser, exclude=['-h', '--help'])
args = parser.parse_args()
```
Expected:
```
$ ./test.py <tab>
--test --no_test
$ ./test.py --test <tab>
```

I could not found a solution here, when I used my code (PR #143), I got close, but because it's exclusive, if I chose --test no I will get forever --no_test on the compilation because it's always False.

evanunderscore:
Fixes #144

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -336,9 +336,20 @@ class CompletionFinder(object):
         for action in parser._actions:
             if action.help == argparse.SUPPRESS and not self.print_suppressed:
                 continue
+            if not self._action_allowed(action, parser):
+                continue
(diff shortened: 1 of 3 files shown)


kislyuk:
Great, thank you.

PR #147, opened 2016-10-13
Fix headers consistency in the Changes.rst

svetlyak40wt:
Last two versions were made higher level headers than pervious. This inconsistent and breaks automatic changelog parsing at https://allmychanges.com/p/python/argcomplete/


kislyuk:
If you wish to make this consistent, please replace all the old headers instead. These headers were added by the pandoc auto-conversion in the release make target.

svetlyak40wt:
@kislyuk done!

kislyuk:
Thanks! :)

Issue #20, opened 2013-03-16
problems with mixed optional and positional arguments


jimparis:
With this test program:

```
#!/usr/bin/env python
import argparse
import argcomplete
def c_test(prefix, parsed_args, **kwargs):
    return [ "name1", "name2" ]
parser = argparse.ArgumentParser()
parser.add_argument("name").completer = c_test
exc = parser.add_mutually_exclusive_group()
exc.add_argument("--get", action="store_true")
exc.add_argument("--set", action="store_true")
argcomplete.autocomplete(parser)
args = parser.parse_args()
```

Usage is listed as:

```
usage: test.py [-h] [--get | --set] name
```

But optional arguments aren't suggested as completions, and completions for `name` are listed even after a `name` has been supplied.

Expected:

```
$ ./test.py <tab>
-h --get --set name1 name2
$ ./test.py --<tab>
--get --set
$ ./test.py --get <tab>
-h name1 name2
$ ./test.py --get name1 <tab>
-h
```

Actual:

```
$ ./test.py <tab>
name1 name2
$ ./test.py --<tab>
(nothing)
$ ./test.py --get <tab>
name1 name2
$ ./test.py --get name1 <tab>
name1 name2
```

PR #148, opened 2016-10-14
Add test case to verify #20 is fixed

evanunderscore:


evanunderscore:
Not sure how necessary this is, but I didn't see any existing tests that handled this particular combination. Happy for this to just be closed if you don't think it adds any value. Either way, #20 can be closed.

kislyuk:
Can't say I know whether this adds coverage, but it doesn't hurt to have this test. Thanks.

PR #149, opened 2016-10-14
Fix parser reuse with positional arguments

evanunderscore:
Minimal changes to fix the expected failure test case for reusing a parser. The problem was `num_consumed_args` was never reset on the actions. Since I don't know of an easy way to reset all the actions, I've changed this to instead store that count in a global that is reset when arguments are parsed. 

A more natural way to do this would be to attach the dictionary to the parser, but that would require making the three `action_is_*` functions be methods of `IntrospectiveArgumentParser`.


kislyuk:
Great, thanks!

PR #150, opened 2016-10-15
Fix parser reuse

evanunderscore:
In a REPL environment, parsers and subparsers continued to use the patched `_parse_known_args` method forever. This adds a similar check to the one already present in `IntrospectAction` to fall back to the base implementation when not completing.

### DIFF
--- a/argcomplete/my_argparse.py
+++ b/argcomplete/my_argparse.py
@@ -59,7 +59,6 @@ def action_is_greedy(action, isoptional=False):
 class IntrospectiveArgumentParser(ArgumentParser):
     ''' The following is a verbatim copy of ArgumentParser._parse_known_args (Python 2.7.3),
     except for the lines that contain the string "Added by argcomplete".
-    Lines that contain the string "Disabled by argcomplete" have been commented out.
     '''
 
     def _parse_known_args(self, arg_strings, namespace):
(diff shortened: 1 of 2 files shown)


evanunderscore:
Instead of cd8c83c21f8da26c163c68637c284c8264eb7c0a, it seems this can also be fixed by simply uncommenting the code that was commented out in 02ccecaa98f90be5c6caf4622b83c40dfee070af (the test case that commit added still passes).

It looks like the only difference that makes is whether `CompletionFinder._get_completions` is able to print the parsed arguments or not. Should we just uncomment that code instead? Should we do both?

kislyuk:
I would prefer going with the second solution (uncommenting the code that was commented out in 02cceca) - it minimizes departure from upstream - but I'm surprised the test case that commit added still passes. There must have been a reason @dankilman commented out that block, I was under the impression it was necessary in incomplete parse situations.

dankilman:
To be honest, I don't remember what made me comment out that section, I most certainly gave it some thought at the time but can't recall it now. It is indeed strange to me that the test cases I added still pass

evanunderscore:
The changes seem sensible, but since the namespace is populated as it goes and the return value is effectively thrown away, I can't see why it would matter whether it fails there.

kislyuk:
It's unclear to me whether we want to make a full commitment to making sure the parser is functional after completion runs (the REPL compatibility feature notwithstanding - do many people really use it?) - but it seems like it won't be too hard to do so. Merging.

PR #152, opened 2016-10-16
Don't insert leading quote char

evanunderscore:
Bash doesn't appear to need this; it only clutters up the completion display.

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -511,7 +511,7 @@ class CompletionFinder(object):
             if char not in comp_wordbreaks:
                 comp_wordbreaks += char
 
-        # If the word under the cursor was quoted, escape the quote char and add the leading quote back in.
+        # If the word under the cursor was quoted, escape the quote char.
         # Otherwise, escape all COMP_WORDBREAKS chars.
         if cword_prequote == "":
             # Bash mangles completions which contain colons if COMP_WORDBREAKS contains a colon.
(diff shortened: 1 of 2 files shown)


evanunderscore:
I've noticed the tests don't actually confirm this works in bash, so I'll look at adding some tests with pexpect as in #141.

evanunderscore:
Tests added separately in #153

kislyuk:
I was surprised to see that it all works without the quote. Hopefully we didn't miss anything. One thing to add to the tests is completions with multiple special characters, like a:b:c or a$b$c (the latter still causes problems in all cases, I think).

evanunderscore:
Can you remember what problems you had? I added some tests in #154 and the only problem was `$` not being escaped when unquoted.

kislyuk:
```
#!/usr/bin/python
# PYTHON_ARGCOMPLETE_OK
import argparse
import argcomplete
parser.add_argument("--foo", choices=["a:b", "a:c", "b:c", "b:cdef:g", "b:cdef:h", "a$b$c", "a$b$d"], nargs=5)
print(parser.parse_args())
```

Try `./script --foo a$b<TAB>`. It didn't work for me. `./script --foo b:cdef<TAB>` works fine though.

evanunderscore:
With the change in #154, it works for me with `a\$b`, but `a$b` replaces itself with `b\:c`. I'm not sure what the expected output is, but it's probably not that.

kislyuk:
It's breaking the word at $ and starting the completion with b. I'm not sure if we can do anything about that. Does it work fine with `"a$b<TAB>`?

kislyuk:
Or `'a$b<TAB>`?

evanunderscore:
Those both work for me. The problem with `a$b` is not bash, it's with the way we're splitting the line. Adding `$` to line 77 in `my_shlex` makes this work as intended.

PR #154, opened 2016-10-16
Fix handling of unquoted completions containing $

evanunderscore:


evanunderscore:
The second commit was inspired by the discussion in #152, since the fix for that requires adding `$` to `wordchars`, *not* `punctuation_chars`.

kislyuk:
Could you add tests that round-trip these characters through bash? e.g. `./script a$b$<TAB>` -> `a$b$c`, `a$b$d`

evanunderscore:
It looks like there's a separate issue that stops it from completing correctly escaped dollar signs in double-quotes, e.g. `a\$b<TAB>` breaks. I haven't looked into why this is.

PR #157, opened 2016-10-22
Expose COMP_TYPE environment variable

### LINKED ISSUE #156: Make COMP_TYPE available to completion functions
It would be useful for the `COMP_TYPE` environment variable to be available in a custom completion function, either from `os.environ` or as a keyword argument passed to the function.

This would allow a completion function to respond differently depending on the type of completion, such as differentiating between single tab and double tab completion.

A minimal implementation would only require adding `COMP_TYPE="$COMP_TYPE" \` in two places:

- [argcomplete/bash_completion.d/python-argcomplete.sh#L42](https://github.com/kislyuk/argcomplete/blob/db2c2fa1835b4fb80e90240ea01f04c05426762b/argcomplete/bash_completion.d/python-argcomplete.sh#L42)
- [scripts/register-python-argcomplete#L29](https://github.com/kislyuk/argcomplete/blob/4e278d80735c89346a8b91b4c7a5192b13ef5377/scripts/register-python-argcomplete#L29)

mattclay:
Resolves #156 

Tested both `python-argcomplete.sh` and `register-python-argcomplete` on Ubuntu 16.04 with:
`GNU bash, version 4.3.46(1)-release (x86_64-pc-linux-gnu)`

The first tab completion passes: `COMP_TYPE=9`
The second tab completion passes: `COMP_TYPE=63`

### DIFF
--- a/argcomplete/bash_completion.d/python-argcomplete.sh
+++ b/argcomplete/bash_completion.d/python-argcomplete.sh
@@ -40,6 +40,7 @@ _python_argcomplete_global() {
         COMPREPLY=( $(_ARGCOMPLETE_IFS="$IFS" \
             COMP_LINE="$COMP_LINE" \
             COMP_POINT="$COMP_POINT" \
+            COMP_TYPE="$COMP_TYPE" \
             _ARGCOMPLETE_COMP_WORDBREAKS="$COMP_WORDBREAKS" \
             _ARGCOMPLETE=$ARGCOMPLETE \
             "$executable" 8>&1 9>&2 1>/dev/null 2>&1) )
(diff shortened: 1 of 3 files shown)


kislyuk:
I think the way to test this is by adding an `assert "COMP_TYPE" in os.environ` before `argcomplete.autocomplete()` is called in https://github.com/kislyuk/argcomplete/blob/master/test/prog.

kislyuk:
Great, thanks! I will roll a new release sometime in the next few days.

mattclay:
@kislyuk Is there any chance there will be a release soon that includes this change?

kislyuk:
Released in v1.7.0, sorry about the delay.

Issue #151, opened 2016-10-15
Using '-o nospace' prevents adding spaces after quoted completion


evanunderscore:
Since `quote_completions` doesn't add the trailing space when the input is quoted, nothing adds the space after the closing quote:

```
#!/usr/bin/env python
import argparse
import argcomplete
p = argparse.ArgumentParser()
p.add_argument('foo', choices=['foo', 'bar'])
argcomplete.autocomplete
```

```
$ test f<tab>
$ test foo <cursor>

$ test "f<tab>
$ test "foo"<cursor>

$ test 'f<tab>
$ test 'foo'<cursor>
```

This could be fixed by letting bash handle completions as described in #50.

kislyuk:
If there are no side effects like what we found in #50, maybe we could add a space after a quoted completion if it's unique. With that said, having to press space to continue putting together the command line doesn't seem like a huge issue to me in this case.

evanunderscore:
The space would have to be added after the closing quote, which is being inserted by bash. I experimented with getting argcomplete to add the closing quote but bash just adds another one after it so you end up with `"foo" "`.

Another option is to offload the nospace decision to the bash script in a similar fashion to http://stackoverflow.com/questions/8503126/custom-directory-completion-appends-whitespace/12185493#12185493.

```
if [[ "$COMPREPLY" =~ [/=:]$ ]]; then
    compopt -o nospace
fi
```

This would mean we can always return completions without spaces, which may also solve #50.

evanunderscore:
This was fixed in #159

PR #159, opened 2016-10-22
Let bash add or suppress trailing space

### LINKED ISSUE #50: zsh escapes space after singleton completions

evanunderscore:
Fixes #50 and #151

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -106,7 +106,7 @@ class CompletionFinder(object):
     :meth:`CompletionFinder.__call__()`.
     """
     def __init__(self, argument_parser=None, always_complete_options=True, exclude=None, validator=None,
-                 print_suppressed=False, default_completer=FilesCompleter()):
+                 print_suppressed=False, default_completer=FilesCompleter(), append_space=None):
         self._parser = argument_parser
         self.always_complete_options = always_complete_options
         self.exclude = exclude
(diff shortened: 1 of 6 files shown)


evanunderscore:
I know this may be somewhat controversial since it's a fairly fundamental shift in the interaction between argcomplete and the shell. I'm also not aware of exactly how this will affect shells like zsh or tcsh.

kislyuk:
Thanks for the PR. I think I'm on board with this, but I'm still trying to understand the possible side effects (or lack thereof) of this.

Could you please rebase this on master so it's good to merge?

evanunderscore:
As far as I can tell, there should be no difference when this is used from bash. However this will definitely impact anyone using the completions directly since they'll be expecting exact matches to have a space appended, so in that sense this is a backwards compatibility break. If you'd prefer, we could make this opt-in for now.

evanunderscore:
I've updated this to account for the testing changes. In my opinion, merging this would require a major version bump since some code may rely on the trailing space that is being removed (assuming you're doing semver). Alternatively I could make it optional and have the bash scripts activate this change through an environment variable or some other mechanism.

evanunderscore:
On reflection I think it's a better idea to make this optional. I'll make the appropriate changes.

evanunderscore:
I've updated this so that the default remains unchanged and environments that are prepared to handle this for themselves can opt out. This is ready for another review.

kislyuk:
Looks good. Merging.

I'll play around with this and once I understand the implications (or lack thereof), I'll put my opinion here on whether it should be on by default.

PR #182, opened 2017-01-10
Only use compopt when it is supported

### LINKED ISSUE #181: Continuation characters display error on Bash 3.x
An error is printed in Bash 3.x when a completion is returned that ends in a continuation character. [This test](https://github.com/kislyuk/argcomplete/blob/a433642df8aaac49e24be1fbda640f5777acd03e/test/test.py#L849) fails as a result.

The source of this problem is that compopt is not available in 3.x. The options are:
* Suppress the error (this won't stop the space being inserted)
* Revert #159 
* Add another option when registering, e.g. `--shell bash3` that restores the behavior before #159

evanunderscore:
Reverts changes to `complete_options` in #159, fixes #181.


kislyuk:
LGTM

PR #187, opened 2017-01-13
Specially handle all characters in COMP_WORDBREAKS

### LINKED ISSUE #94: Does argcomplete special case = (equals signs)?
Does argcomplete special case equals signs, or am I doing this wrong?

For conda, I want to complete `conda install packagename=version`, like `conda install python=3.3`. I am using

``` py

# To get tab completion from argcomplete
class Completer(object):
    @memoize
    def get_items(self):
        return self._get_items()

    def __contains__(self, item):
        # This generally isn't all possibilities, and even if it is, we want
        # to give better error messages than argparse
        return True

    def __iter__(self):
        return iter(self.get_items())

class Packages(Completer):
    def __init__(self, prefix, parsed_args, **kwargs):
        self.prefix = prefix
        self.parsed_args = parsed_args

        self.has_eq = '=' in prefix

    def _get_items(self):
        # TODO: Include .tar.bz2 files for local installs.
        from conda.api import get_index
        args = self.parsed_args
        index = get_index(channel_urls=args.channel or (), use_cache=True,
            prepend=not args.override_channels, unknown=args.unknown,
            offline=args.offline)
        if not self.has_eq:
            return [i.rsplit('-', 2)[0] for i in index]
        else:
            return [i.rsplit('-', 2)[0] + '=' + i.rsplit('-', 2)[1] for i in index]
```

Sorry that the code is a little complicated. I have a lot of completers. 

The idea is that if the package doesn't contain a `=`, we just return a list of packages (because the common case is to not include any `=` at the command line, like `conda install numpy`). If it does, I want to complete to `package=version`. But this is giving

```
$PYTHONPATH=. conda install numpy= # I press TAB here
numpy\=1.5.1     numpy\=1.6.2     numpy\=1.7.0     numpy\=1.7.0b2   numpy\=1.7.0rc1  numpy\=1.7.1     numpy\=1.8.0     numpy\=1.8.1     numpy\=1.8.2     numpy\=1.9.0     numpy\=1.9.1
$PYTHONPATH=. conda install numpy=numpy\=1.
```

(the PYTHONPATH is to make sure I am using the git master of argcomplete to test if this was already fixed)

How do I get the behavior I want?

evanunderscore:
Fixes #94

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -27,6 +27,7 @@ else:
     import unittest2 as unittest
 
 IFS = "\013"
+COMP_WORDBREAKS = " \t\n\"'><=;&(:"
 
 BASH_VERSION = subprocess.check_output(['bash', '-c', 'echo $BASH_VERSION']).decode()
 BASH_MAJOR_VERSION = int(BASH_VERSION.split('.')[0])
@@ -63,6 +64,7 @@ class TestArgcomplete(unittest.TestCase):
         os.environ["_ARGCOMPLETE"] = "yes"
         os.environ["_ARC_DEBUG"] = "yes"
         os.environ["IFS"] = IFS
+        os.environ["_ARGCOMPLETE_COMP_WORDBREAKS"] = COMP_WORDBREAKS
 
     def tearDown(self):
         os.environ = self._os_environ
@@ -76,7 +78,6 @@ class TestArgcomplete(unittest.TestCase):
         with TemporaryFile() as t:
             os.environ["COMP_LINE"] = ensure_bytes(command) if USING_PYTHON2 else command
             os.environ["COMP_POINT"] = point
-            os.environ["_ARGCOMPLETE_COMP_WORDBREAKS"] = '"\'@><=;|&(:'
             self.assertRaises(SystemExit, completer, parser, output_stream=t,
                               exit_method=sys.exit, **kwargs)
             t.seek(0)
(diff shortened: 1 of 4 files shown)


evanunderscore:
Important things:

* This breaks quotes in tcsh, but they can be fixed fairly easily
* COMP_WORDBREAKS no longer has a default, which may impact anyone using this outside of bash
* This will make some things display weirdly, like the URLs in the tests

```
$ ./prog url h<tab>
$ ./prog url http://url<tab>
//url1  //url2  
```

In the above example, completion functions correctly, but printing is trimmed at the colon. Previously this happened sometimes, depending on whether the user had typed an unescaped colon:

```
$ ./prog url h<tab>
$ ./prog url http\://url<tab>
http\://url1  http\://url2
$ ./prog url http:<tab>
$ ./prog url http://url<tab>
//url1  //url2
```

evanunderscore:
@kislyuk Are you okay with these changes, noting the differences I've highlighted above?

kislyuk:
What are the characters for which truncation of printed suggestions happens?

evanunderscore:
Anything in `COMP_WORDBREAKS` that isn't in the list of characters we're always escaping. For what seems to be the default value of that, this will be `=` and `:`.

test/test.py
```diff
@@ -943,6 +954,15 @@ def test_single_quotes_in_single_quotes(self):
...
         self.assertEqual(self.sh.run_command("prog quote '1\t"), "1'1\r\n")
 
+    def test_wordbreak_chars(self):
+        self.assertEqual(self.sh.run_command('prog break a\tc'), 'a:b:c\r\n')
+        self.assertEqual(self.sh.run_command('prog break a:b:\tc'), 'a:b:c\r\n')
+        self.assertEqual(self.sh.run_command('prog break a:b\tc'), 'a:b:c\r\n')
+        self.assertEqual(self.sh.run_command("prog break 'a\tc'"), 'a:b:c\r\n')
+        self.assertEqual(self.sh.run_command("prog break 'a:b\tc\t"), 'a:b:c\r\n')
+        self.assertEqual(self.sh.run_command('prog break "a\tc"'), 'a:b:c\r\n')
+        self.assertEqual(self.sh.run_command('prog break "a:b\tc\t'), 'a:b:c\r\n')
```
kislyuk: What are the double-tab tests supposed to mean? Do I understand correctly that the `\t` means pexpect will send TAB to the subprocess and trigger completion? What exactly do the characters after `\t` mean then? Aren't they disregarded and never passed to the completer? And shouldn't two tabs trigger completion twice? Thanks.

kislyuk:
LGTM although I confess I don't quite understand the pexpect-based tests as per my comment.

test/test.py
evanunderscore: Yes, the tests with two tabs will trigger completion twice. For example, the last test is simulating the user doing the following:

1. Type `prog break "a:b`
1. Press TAB, completer returns `['a:b:c', 'a:b:d']`, shell fills in `:`
1. Type `c`
1. Press TAB, completer returns `['a:b:c']`, shell fills in `" `
1. Press enter, submitting `prog break "a:b:c" `

evanunderscore: I've added a similar comment to the base test class. Does that make more sense?

evanunderscore:
Thanks for reviewing.

kislyuk:
Great, thanks for adding the docstring, that's very informative.

PR #195, opened 2017-01-18
Simplify nospace handling in global completion

evanunderscore:
I had previously changed this because `compopt` isn't available in older versions of Bash, but neither is global completion, so there's no need for the workaround here. This change means it's not necessary to remember to call `compopt +o nospace` before returning to get default completion.


kislyuk:
LGTM

Issue #196, opened 2017-01-19
wordbreak char completion issues


kislyuk:
@evanunderscore, I'm filing this to run it by you in case you expect it to work differently. I'm not entirely sure what to expect with all the recent changes.

### Test case
```
#!/usr/bin/python
# PYTHON_ARGCOMPLETE_OK
import argparse, argcomplete
parser = argparse.ArgumentParser()
parser.add_argument("--foo", choices=["a:b", "a:c", "b:c", "b:cdef:g", "b:cdef:h", "a$b$c", "a$b$d"], nargs=5)
argcomplete.autocomplete(parser)
print(parser.parse_args())
```
### Completion run 1
```
$ prog --foo b:cdef:g<TAB>
```
```
   g  h
```
This displays two non-functioning completions with some spaces prepended, when it should add a space to finalize the completion and move on.
### Completion run 2
```
$ prog --foo "a$<TAB>
$ prog --foo "a\$b\$<TAB>
```
This wedges the completer, while it should produce `a\$b\$c a\$b\$d` or `"a$b$c "a$b$d`.

kislyuk:
One more:
### Completion run 3
```
$ prog --foo a:b <TAB>
b       b\$c    b\$d    c       cdef:g  cdef:h
```
This produces completions as if `a:<TAB>` was typed, while it should produce what `prog --foo <TAB>` gives:
```
a:b       a:c       a\$b\$c   a\$b\$d   b:c       b:cdef:g  b:cdef:h
```

kislyuk:
If I run against v1.7.0, cases 1 and 3 are fixed, but case 2 is an issue as well.

evanunderscore:
1. Definitely looks odd, but I can't reproduce it using master and global completion. Can you include some information about your environment? Shell version, Python version, value of `COMP_WORDBREAKS` should help. Also whether you're using global completion, and the output of `type _python_argcomplete_global` to be sure you have the right script.
2. This is an expected failure, see [here](https://github.com/kislyuk/argcomplete/blob/19c6544c303da72d16985312b3f8d15e205ca4f4/test/test.py#L955)
3. This is expected behavior, see discussion on #187

evanunderscore:
For the 1.8.0 release notes, you forgot to credit the authors of the merge requests that weren't me (#167, #143).

evanunderscore:
Sorry, I misread 3. That is definitely a bug, and I'm able to reproduce it on my machine. I'll look into it.

[...]

PR #197, opened 2017-01-20
Fix completion after tokens with wordbreak chars

evanunderscore:
Fixes part of #196


kislyuk:
Fix confirmed, thanks!

PR #201, opened 2017-01-23
Clarify role of prefix in completers

### LINKED ISSUE #139: Clarify that filtering by prefix is not required in completers

evanunderscore:
Fixes #139

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -140,9 +140,9 @@ organization and complete their names, then prints the member description:
     # PYTHON_ARGCOMPLETE_OK
     import argcomplete, argparse, requests, pprint
 
-    def github_org_members(parsed_args, **kwargs):
+    def github_org_members(prefix, parsed_args, **kwargs):
         resource = "https://api.github.com/orgs/{org}/members".format(org=parsed_args.organization)
-        return (member['login'] for member in requests.get(resource).json())
+        return (member['login'] for member in requests.get(resource).json() if member['login'].startswith(prefix))
 
     parser = argparse.ArgumentParser()
     parser.add_argument("--organization", help="GitHub organization")
--- a/README.rst
+++ b/README.rst
@@ -75,7 +75,8 @@ Specifying completers
 You can specify custom completion functions for your options and arguments. Two styles are supported: callable and
 readline-style. Callable completers are simpler. They are called with the following keyword arguments:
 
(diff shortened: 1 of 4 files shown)


evanunderscore:
@kislyuk I've altered some of the documentation and examples to remove the use of the prefix as a filter. I've also change the actual completers to do the same and be generally simpler and more lazy. Happy to undo some of these changes if you don't think they're necessary.

docs/examples/describe_github_user.py
```diff
@@ -2,9 +2,9 @@
 # PYTHON_ARGCOMPLETE_OK
 import argcomplete, argparse, requests, pprint
 
-def github_org_members(prefix, parsed_args, **kwargs):
+def github_org_members(parsed_args, **kwargs):
     resource = "https://api.github.com/orgs/{org}/members".format(org=parsed_args.organization)
-    return (member['login'] for member in requests.get(resource).json() if member['login'].startswith(prefix))
+    return (member['login'] for member in requests.get(resource).json())
```
kislyuk: I want to retain this in the example. Please don't remove it.

README.rst
```diff
@@ -139,9 +140,9 @@ organization and complete their names, then prints the member description:
     # PYTHON_ARGCOMPLETE_OK
     import argcomplete, argparse, requests, pprint
 
-    def github_org_members(prefix, parsed_args, **kwargs):
+    def github_org_members(parsed_args, **kwargs):
         resource = "https://api.github.com/orgs/{org}/members".format(org=parsed_args.organization)
-        return (member['login'] for member in requests.get(resource).json() if member['login'].startswith(prefix))
+        return (member['login'] for member in requests.get(resource).json())
```
kislyuk: Same here, let's retain this.

kislyuk:
Looks OK modulo the two comments, let's keep the examples.

evanunderscore:
I'll revert those two.

Perhaps if we added another example that uses the prefix they could be removed again? Something like #94 looks like a good candidate for this, e.g.:

```
if '=' not in prefix:
    return get_names()
name, _ = prefix.split('=')
return (f'{name}={v}' for v in get_versions(name))
```

As it stands the example gives the impression that filtering is a useful thing for a completer to do.

evanunderscore:
I've reverted the changes to the example script. I'll merge this in. Thanks for the review.

PR #202, opened 2017-01-24
Increase timeout for slow build machine

evanunderscore:


evanunderscore:
@kislyuk Let me know if you see this failure again and I'll take another look.

PR #203, opened 2017-01-24
Move timeout in tests

evanunderscore:
Revised attempt for #202


evanunderscore:
I've rerun this job a handful of times without seeing the failure, so hopefully this is fixed properly now.

PR #204, opened 2017-01-27
Complete python -m module

### LINKED ISSUE #173: Support for python -m module
Is it possible for argcomplete to work for modules run as `python -m module`? This is necessary to do once you start organizing your code in a module (rather than just a single script) to make relative imports work correctly, but I don't really want to make a separate entry point script if I don't have to. 

I am using global completion and it isn't working, at least not out of the box.

evanunderscore:
Fixes #173


kislyuk:
LGTM although the "_scripts" subdirectory seems superfluous, you could just put the script at package top level.

asmeurer:
I tried testing this and it didn't work. I checked out your branch and ran

```
export PYTHONPATH=~/Documents/argcomplete/
```

and completions did not work for `python -m mymodule`.

asmeurer:
Oh nevermind, I forgot to resource the completion script. 

It seems to work. Would be possible to complete the module names after `-m` as well?

kislyuk:
I dislike this feature creep. What you describe is a job for a dedicated python completer.

evanunderscore:
I would like it to be able to complete module names, but that could be difficult, and probably doesn't belong in this PR.

I've removed the `_scripts` directory. I'll merge this in.

kislyuk:
Great, thanks.

Module completions seem to be available upstream as of https://github.com/scop/bash-completion/commit/0e8d34e, which is from 2013 and says it's available in 2.1. I have bash-completion-1.3_1 installed though, and `python -m<TAB>` doesn't seem to be working. Is this an upstream bug?

kislyuk:
D'oh, of course, 1.3 is older than 2.1. It looks like Homebrew's version is way out of date.

Looking at http://packages.ubuntu.com/search?keywords=bash-completion, 2.1 is available as of Ubuntu 14.04+. But on my 16.04 instance, `python -m<TAB>` doesn't seem to be working either.

kislyuk:
OK, it works if I do `source /usr/share/bash-completion/completions/python`. So it looks like you have to turn it on manually by running `ln -s /usr/share/bash-completion/completions/python /etc/bash_completion.d` or something.

evanunderscore:
I have 2.1 but no python in the completions folder, so I'm not sure what that's about. The default completion function is supposed to source that file for you the first time you attempt to complete `python`. If you have the argcomplete's global completion installed, it's going to steal completion away for python before it gets to that.

evanunderscore:
I've created #207 to capture the problem of global completion not interacting with bash-completion.

PR #205, opened 2017-01-27
Clarify supported Python versions

evanunderscore:


README.rst
```diff
@@ -15,6 +15,8 @@ the network).
 
 Installation
 ------------
+Argcomplete requires Python 2.7 or 3.3+.
```
kislyuk: Please move this to the bottom of the readme, next to the *Common Problems* section.

Space at the top of the readme is precious and we shouldn't clutter it with warnings about python versions that roughly nobody uses.

evanunderscore: I moved it under its own heading, but I don't know if it looks weird because the section is so short.

I also removed another bit of 2.6-specific code and updated the RTD link.

evanunderscore:
Merging - if you don't like the way the readme turned out, let me know and I'll change it.

kislyuk:
LGTM, thanks.

PR #206, opened 2017-01-27
Redirect all output to debug stream in debug mode

### LINKED ISSUE #97: debug mode should not mute output
Debug mode shouldn't mute the output, especially stderr. Usually when I get things wrong, it's because some traceback is raised (e.g., an AttributeError in the parsed_args argument of the completer). 

I tried disabling the muting in debug mode, but it didn't work. I'm not sure what I was doing wrong.

evanunderscore:
Fixes #97

### DIFF
--- a/argcomplete/bash_completion.d/python-argcomplete.sh
+++ b/argcomplete/bash_completion.d/python-argcomplete.sh
@@ -12,6 +12,17 @@ __python_argcomplete_expand_tilde_by_ref () {
     fi
 }
 
+# Run something, muting output or redirecting it to the debug stream
+# depending on the value of _ARC_DEBUG.
+__python_argcomplete_run() {
+    if [[ -z "$_ARC_DEBUG" ]]; then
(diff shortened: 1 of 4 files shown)


argcomplete/bash_completion.d/python-argcomplete.sh
```diff
@@ -12,6 +12,17 @@ __python_argcomplete_expand_tilde_by_ref () {
...
 
+# Run something, muting output or redirecting it to the debug stream
+# depending on the value of _ARC_DEBUG.
+__python_argcomplete_run() {
+    if [[ -z "$_ARC_DEBUG" ]]; then
+        "$@" 8>&1 9>&2 1>/dev/null 2>&1
+    else
+        "$@" 8>&1 9>&2 1>&9 2>&1
+    fi
+    return $?
```
kislyuk: Doesn't this happen by default?

evanunderscore: I wasn't aware of that. Thanks.

PR #211, opened 2017-02-28
Fixes #209 "Bad file descriptor" under unit test.

### LINKED ISSUE #209: "Bad file descriptor" under unit test
Unit tests which repeatedly exercise argcomplete in the output contain multiple occurrences of:
```
  close failed in file object destructor:
  IOError: [Errno 9] Bad file descriptor
```

which seems to be related to
```
       global debug_stream
       try:
           debug_stream = os.fdopen(9, 'w')
       except:
           debug_stream = sys.stderr
```

If the above is guarded by
   `if _DEBUG:`
where (`_DEBUG = '_ARC_DEBUG' in os.environ`) the message disappears.

cherba:


kislyuk:
No, this solution won't work, because the debug stream is also used in the `warn()` function - please see the README for the doc on how it's used.

cherba:
Thanks for looking at propose PR. Can you elaborate a little more on why this is an issue.

> debug stream is also used in the warn()
yes, but this change does not remove debug_stream it would still be set to stderr as debug_stream is initialized right from the start.

Also does
   debug_stream = os.fdopen(9, "w")
make sense outside _DEBUG mode? I'd be happy to add any other workaround.

PR #218, opened 2017-04-17
Fix test for non-English locales

### LINKED ISSUE #217: TestBashGlobal::test_python_not_executable fails with non-english locale
The test fails at [this line](https://github.com/kislyuk/argcomplete/blob/master/test/test.py#L1102).

For example, in a German locale the error message returned by `self.sh.run_command` is `'bash: ./prog: Keine Berechtigung\r\n'` and of course doesn't contain the string the test looks for.

Setting `LANG=en_US.UTF-8` or even `LC_ALL=C` somehow should fix this.

evanunderscore:
Fixes #217


kislyuk:
LGTM

PR #224, opened 2017-08-15
add SuppressCompleter to skip completion for specific arguments while allowing help text

### LINKED ISSUE #223: Skip completion for specific argument
I would like a specific argument (e.g. `--foo`) to not appear in the completion. Since the application is pretty modular I can't configure e.g. the global `CompletionFinder`. I found the option to set `help=argparse.SUPPRESS` but I still want a help text to be shown for `--help`.

I was thinking if something like this would be feasible:

```
arg = parser.add_argument('--foo', help='awesome option')
arg.completer = False
```

If anyone could be give me some pointers I am happy to provide a PR for this feature idea.

dirk-thomas:
Fixes #223.

The first commit adds the `SuppressCompleter` as described in the feature request.

The second commit adds a method to it which makes the actual decision if the completion should be suppressed. The default implementation is to suppress the completion. But it allows to override the decision making in a subclass and use whatever reasoning to decide to show or suppress the completion.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -135,7 +135,29 @@ class TestArgcomplete(unittest.TestCase):
             parser = ArgumentParser()
             parser.add_argument("--foo")
             parser.add_argument("--bar", help=SUPPRESS)
-            arg = parser.add_argument("--baz")
+            return parser
+
+        expected_outputs = (
+            ("prog ", ["--foo", "-h", "--help"]),
+            ("prog --b", [""])
+        )
+
+        for cmd, output in expected_outputs:
+            self.assertEqual(set(self.run_completer(make_parser(), cmd)), set(output))
+
+        expected_outputs = (
+            ("prog ", ["--foo", "--bar", "-h", "--help"]),
+            ("prog --b", ["--bar "])
+        )
+
+        for cmd, output in expected_outputs:
+            self.assertEqual(set(self.run_completer(make_parser(), cmd, print_suppressed=True)), set(output))
+
+    def test_suppress_completer(self):
+        def make_parser():
+            parser = ArgumentParser()
+            parser.add_argument("--foo")
+            arg = parser.add_argument("--bar")
             arg.completer = SuppressCompleter()
             return parser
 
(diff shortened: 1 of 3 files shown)


test/test.py
```diff
@@ -146,8 +148,8 @@ def make_parser():
             self.assertEqual(set(self.run_completer(make_parser(), cmd)), set(output))
 
         expected_outputs = (
-            ("prog ", ["--foo", "--bar", "-h", "--help"]),
-            ("prog --b", ["--bar "])
+            ("prog ", ["--foo", "--bar", "--baz", "-h", "--help"]),
+            ("prog --b", ["--bar", "--baz"])
```
dirk-thomas: I am not sure why I had to remove the trailing space for the existing option `--bar`?

dirk-thomas:
The first commit add the SuppressCompleter as described in the feature request.

The second commit adds a method to it which makes the actual decision if the completion should be suppressed. The default implementation is to suppress the completion. But it allows to override the decision making in a subclass and use whatever reasoning to decide to show or suppress the completion.

kislyuk:
Please create a separate test instead of modifying an existing test.

Please add a docstring to `SuppressCompleter`. I'll make sure it makes its way into the autodoc (I think that file might not currently be indexed by autodoc).

kislyuk:
The lack of a trailing space is suspicious. By default we add a trailing space if there is only one completion, but it seems like that shouldn't change in your test. Once you refactor the test case to be its own test case, I'll take a closer look into what's going on there.

dirk-thomas:
> Please create a separate test instead of modifying an existing test.

I thought it would make sense to put the additional check into the similar existing test. I have separated it in 89c3a0ef21698d2a01a4dddb169810d9aa13400b.

> Please add a docstring to SuppressCompleter.

Done in 1b3629e22c96d46f3b39bc84133486c0f2a36ce4

> The lack of a trailing space is suspicious.

With the separation the tests I didn't have to change the existing test and also the new test uses the trailing space.

---

I could squash the third commit with the first one  as well as the fourth with the second if you would like to keep a clean commit history?

kislyuk:
Thanks. That's all right, I can squash merge when I'm done reviewing, unless you really want me to rebase merge.

evanunderscore:
The tests don't show what happens when the user types `prog --bar <tab>`. What should happen in this case?

dirk-thomas:
> The tests don't show what happens when the user types `prog --bar <tab>`. What should happen in this case?

For the specific test where one argument is expected after `--bar` I wouldn't expect any completion. If `--bar` would be changed to `action='action_true'` I would expect "normal" completion of the remaining options to be shown.

dirk-thomas:
@kislyuk @evanunderscore Please let me know if there is anything else I can do for this patch. Thanks.

kislyuk:
Looks good, I verified that there was no problem with the trailing space. Thanks for your patience @dirk-thomas.

kislyuk:
Released in v1.9.2.

PR #226, opened 2017-09-02
Use the correct interpreter when checking wrappers

### LINKED ISSUE #11: Use the Python interpreter specified in the easyinstall wrapper script
`python-argcomplete-check-easy-install-script` should be run using the same interpreter as the wrapper itself, otherwise it may not find the target package (or find the wrong version) due to include path and version differences

### LINKED ISSUE #101: Use of python-argcomplete-check-easy-install-script
I'm using python-argcomplete-check-easy-install-script to check if a setup.py (distutils) script wrapper will work. I'm getting this result: 

```
pkg_resources.DistributionNotFound: dna==0.0.0
```

Here's the wrapper script:

``` python
#!/opt/dna/bin/python
# EASY-INSTALL-SCRIPT: 'dna==0.0.0','dna'
__requires__ = 'dna==0.0.0'
import pkg_resources
pkg_resources.run_script('dna==0.0.0', 'dna')
```

One possible hitch is that I'm using virtualenv.py to install the Python runtime under /opt/dna.

More broadly though - how might I approach debugging this problem?

### LINKED ISSUE #112: Does this work with setuptool's entry_points ?
I've setup:
`mypackage` with a setup.py like so:

```
from setuptools import setup, find_packages
setup(
    name = "HelloWorld",
    version = "0.1",
    packages = find_packages(),
    entry_points={
        'console_scripts': [
            'my-util = mymodule.util:main',
        ]
    }
)
```

I've installed in on a python 2.7 virtualenv and made `/dev/mypackage/.env` my environemnt
`python -m virtualenv .env`
`source .env/bin/activate`
Install mypackage in develop mode:
`python -m pip develop mypackage`

now the `my-util` command has been added to the virtualenv path:
`/dev/mypackage/.env/bin/my-util`

```
#!/dev/mypackage/.env/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'mypackage==0.1','console_scripts','my-util'
__requires__ = 'mypackage==0.1'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('mypackage==0.1', 'console_scripts', 'my-util')()
    )
```

`/dev/mypackage/util/__init__.py`

```
#!/usr/bin/env python
# PYTHON_ARGCOMPLETE_OK
"""
My docstring
"""

import argparse
import os

ARGP = argparse.ArgumentParser(
    description=__doc__,
    formatter_class=argparse.RawTextHelpFormatter,
)
class Completer(object):
    @staticmethod
    def config(prefix, **kwargs):
        return (v for v in os.environ if v.startswith(prefix))
    @staticmethod
    def profile(prefix, **kwargs):
        return (v for v in os.environ if v.startswith(prefix))
ARGP.add_argument('--config', help='`.conf` File to Read.', default='~/mypackage.conf')\
.completer = Completer.config
ARGP.add_argument('profile', nargs='?', help='Profile Name')\
.completer = Completer.profile
argcomplete.autocomplete(ARGP)

from pprint import pprint

def main(argp=None):
    if argp is None:
        argp=ARGP.parse_args()

    print(__doc__)
    pprint(argp, width=120) 
```

---

Am I doing anything obviously wrong? I refresh my bash shell session, reactivate the virtualenv and can tab complete the my-util on the path like normal, but it doesn't autocomplete anything special from there.

evanunderscore:
Fixes #11, possibly also fixes #101 and fixes #112.

### DIFF
--- a/argcomplete/bash_completion.d/python-argcomplete.sh
+++ b/argcomplete/bash_completion.d/python-argcomplete.sh
@@ -47,7 +47,8 @@ _python_argcomplete_global() {
         if (head -c 1024 "$SCRIPT_NAME" | grep --quiet "PYTHON_ARGCOMPLETE_OK") >/dev/null 2>&1; then
             local ARGCOMPLETE=1
         elif (head -c 1024 "$SCRIPT_NAME" | egrep --quiet "(PBR Generated)|(EASY-INSTALL-(SCRIPT|ENTRY-SCRIPT|DEV-SCRIPT))" \
-            && python-argcomplete-check-easy-install-script "$SCRIPT_NAME") >/dev/null 2>&1; then
+            && [[ "$(head -n 1 "$SCRIPT_NAME")" =~ ^#!(.*)$ ]] && [[ "${BASH_REMATCH[1]}" =~ ^.*(python|pypy)[0-9\.]*$ ]] \
+            && "$BASH_REMATCH" "$(which python-argcomplete-check-easy-install-script)" "$SCRIPT_NAME") >/dev/null 2>&1; then
             local ARGCOMPLETE=1
         fi
     fi


evanunderscore:
@kislyuk Are you okay with merging this?

kislyuk:
What is BASH_REMATCH?

evanunderscore:
As I understand it, it's the value of the last thing you matched using a regular expression, and you can index it to get the match groups. The first regex gets the value after the `#!`, the second regex verifies that's some kind of `python` or `pypy` interpreter, then it uses that value to run the script.

evanunderscore:
I couldn't find a good way to comment this with the way the code is structured currently. If you like I can break this logic out into its own function, or I can just merge the way it is now.

kislyuk:
LGTM, sorry about the delay in my review. Upon second reading it makes sense - will just have to be a dense piece of code :)

kislyuk:
@evanunderscore, do you want me to cut a release now, or wait?

Issue #228, opened 2017-09-04
Not work with unicode as input


cuyu:
Example code as below:
```python
def MyCompleter(prefix, **kwargs):
    results = ['aaa', 'bbb']
    return (c for c in results if c.startswith(prefix))

if __name__ == '__main__':
    import argparse
    import argcomplete

    parser = argparse.ArgumentParser()
    subparsers = parser.add_subparsers(dest='command')
    create_parser = subparsers.add_parser('new')
    create_parser.add_argument('title')
    create_parser.add_argument('category').completer = MyCompleter
    argcomplete.autocomplete(parser)
    args = parser.parse_args()
    print args
```

It works as expect when all the inputs are normal characters (assume above code are wrapped in command `demo`):
```bash
$ ./demo.py new 'hi'
--help  -h      aaa     bbb  
```
When the input contains unicode, the auto completer is not working:
```bash
$ ./demo.py new '你好'
```

So, is this a bug or something related to the shell?
I'm using zsh and argcomplete (Version: 1.9.2).

evanunderscore:
It looks like zsh is being smarter than bash and giving COMP_POINT in unicode chars instead of bytes. When we incorrectly interpret that number, we end up slicing the input halfway through a character, which we then fail to decode. Even if that didn't crash, the results would likely be undesirable.

I used bashcompinit as described [here](https://github.com/kislyuk/argcomplete/issues/10#issuecomment-19369876) to reproduce this. I'm assuming bashcompinit is supposed to emulate bash completion, in which case I'd say this should be considered a bug on their end.

On the other hand, [the documentation for COMP_POINT](https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html#Bash-Variables) explicitly says "If the current cursor position is at the end of the current command, the value of this variable is equal to ${#COMP_LINE}." So bashcompinit is adhering to the spec and bash is not.

cuyu:
@evanunderscore Thanks for your quick response.

From my understanding, the problem should only occur on zsh according to your explanation.
However, I just tested on a bash environment, the problem still exists. Can you have a look at this issue on a bash env again?

Thanks!

evanunderscore:
I did test in bash previously and it worked for me. Can you `export _ARC_DEBUG=1`, try again, and tell me if you see a traceback?

cuyu:
Enabled `export _ARC_DEBUG=1` and got the following errors:
```
[...]

PR #229, opened 2017-09-06
Fix bug: crash when inputs contains unicode in zsh

cuyu:
See https://github.com/kislyuk/argcomplete/issues/228 for details.

- Use `ensure_str` can overcome the difference caused by bashcompinit in zsh.
- In python2, the `comp_line ` could be unicode (if the input contains unicode), so add `u` in the debug output:
    - `u'{0}'.format('a')` works
    - `u'{0}'.format(u'你')` works as well
    - `'{0}'.format(u'你')` will fail

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -196,9 +196,9 @@ class CompletionFinder(object):
 
         # Adjust comp_point for wide chars
         if USING_PYTHON2:
-            comp_point = len(comp_line[:comp_point].decode(sys_encoding))
+            comp_point = len(ensure_str(comp_line[:comp_point]))
         else:
-            comp_point = len(comp_line.encode(sys_encoding)[:comp_point].decode(sys_encoding))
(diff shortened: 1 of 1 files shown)


evanunderscore:
Are you sure this fixes the bug for you? The error still happens for me using Python 2 and zsh. Using `ensure_str` amounts to exactly the same thing that was there before.

As far as I can see, there's no way to get around the fact that bash and zsh are giving different values for COMP_LINE, short of detecting which shell we're running in and responding accordingly.

kislyuk:
Neither of the diff hunks in this PR make sense to me (for reasons @evanunderscore described). @cuyu, if you want this PR to proceed, please write a test case that fails without the changes in this PR, and succeeds with them.

cuyu:
Yes, you're right. I misunderstood the root cause of the problem. I will close this PR.

@evanunderscore , @kislyuk Thanks so much for the explanations and comments!

evanunderscore:
@cuyu No problem, thanks for working through this with us.

PR #230, opened 2017-09-06
Fix crash when writing unicode to debug_stream in Python 2

evanunderscore:


cuyu:
I think this commit cannot fix the debug error in https://github.com/kislyuk/argcomplete/issues/228. The problem is actually like the error of below code:
```python
"{l}".format(l=u"你")
```
In this project, the exception is raised from this line, where `comp_line` could be a unicode type:
```python
"\nLINE: '{l}'\nPREQUOTE: '{pq}'\nPREFIX: '{p}'".format(l=comp_line, pq=cword_prequote, p=cword_prefix)
```

evanunderscore:
`argcomplete/__init__.py` imports `unicode_literals` from `__future__` so the `u''` prefix on the strings makes no difference. If you're certain that's where the exception is coming from, I'll need you to help me debug your problem further.

evanunderscore:
Also to be clear, this does not fix the main problem observed in #228; it only fixes the failure to write to the debug stream.

kislyuk:
@evanunderscore, the point of `ensure_bytes()` is you can pass it either strings or bytes, so your inline conditional is unnecessary :)

LGTM otherwise.

evanunderscore:
@kislyuk The arguments are passed straight to `print`, so I was assuming you intended to be able to pass arbitrary objects, which will crash in `ensure_bytes()`. This is what I meant by "non-string input" in the comment.

kislyuk:
OK, sounds good.

evanunderscore:
Thanks for the review. I'll merge this in and hopefully it will make investigating #228 easier.

PR #236, opened 2017-11-14
Fix handling of COMP_POINT

### LINKED ISSUE #228: Not work with unicode as input

evanunderscore:
Fixes #228

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -199,12 +199,6 @@ class CompletionFinder(object):
         comp_line = os.environ["COMP_LINE"]
         comp_point = int(os.environ["COMP_POINT"])
 
-        # Adjust comp_point for wide chars
-        if USING_PYTHON2:
-            comp_point = len(comp_line[:comp_point].decode(sys_encoding))
-        else:
-            comp_point = len(comp_line.encode(sys_encoding)[:comp_point].decode(sys_encoding))
-
         comp_line = ensure_str(comp_line)
         cword_prequote, cword_prefix, cword_suffix, comp_words, last_wordbreak_pos = split_line(comp_line, comp_point)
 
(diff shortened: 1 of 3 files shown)


evanunderscore:
@kislyuk Can I get you to review this please? The code to adjust COMP_POINT has been in there for quite some time, so I want to make sure I'm not missing something important.

kislyuk:
Tested on Python 2 and 3, appears to be correct.

kislyuk:
I'm having difficulty following the details of this behavior on Python 2. On 3, there is no question that it is correct. On 2, I thought the reindexing of comp_point to go from byte coordinates to char coordinates was necessary, but it appears that it's not needed.

kislyuk:
Released in v1.9.3.

evanunderscore:
@kislyuk Thanks for reviewing. As I understand it, it's actually irrelevant whether you're using Python 2 or 3, since Bash determines the value of COMP_POINT by itself based on whatever encoding you're using for your shell. If you set _ARC_DEBUG you can see that COMP_POINT is the same either way, and always corresponds to the length of COMP_LINE in unicode characters.

As I mentioned in the issue, the Bash documentation says COMP_POINT is equal to `${#COMP_LINE}` when you are at the end of the line, and you can test that this works in unicode characters:

```
$ line='prog point test'
$ echo "${#line}"
15
$ line='prog point 你好嘚瑟 '
$ echo "${#line}"
15
```

Based on this, using `ensure_str(os.environ['COMP_LINE'])` seems like it's enough to make everything work as expected.

kislyuk:
Here is why I'm confused:
```
akislyuk@galactica:~/projects/argcomplete>export X=你好嘚瑟; python2 -c'import os; print(len(os.environ["X"]))'
12
akislyuk@galactica:~/projects/argcomplete>export X=你好嘚瑟; python2 -c'import os; print(len(os.environ["X"].decode("utf-8")))'
4
akislyuk@galactica:~/projects/argcomplete>export X=你好嘚瑟; python3 -c'import os; print(len(os.environ["X"]))'
4
```

On Python 3, os.environ contents are automatically decoded for you. If you look at our code, we pass the contents of os.environ["COMP_LINE"] to split_line and then to shlex with no other processing. That means on Python 2, COMP_POINT is being used to index into a bytestring, and on 3, into a unicode string. I'm confused about how that works on Python 2 with wide chars.

kislyuk:
Never mind, I totally missed the part where we run ensure_str. Now it all makes sense.

At some point I must have added ensure_str without thinking about the COMP_POINT reindexing logic (which I should have removed at the same time). So now that's fixed. Thank you.

evanunderscore:
No problem, thanks again for the review.

Issue #235, opened 2017-11-03
Request: Make the shellcode available in a module function


jdevera:
I would like to have my tools to provide their completion shellcode when invoked with a flag, e.g. `--shell-completion`. Right now I have to run `register-python-argcomplete` from my tools, but I would love something more pythonic, like:

```python
print(argcomplete.shellcode(executable='myscript', use_defaults=False, shell='bash'))
```

The the `register-python-argcomplete` script could then even be a thin wrapper around this function.

What do you think?

kislyuk:
Thanks for the suggestion. I'm open to librarizing the shellcode generation like that. Would you like to send a PR?

PR #237, opened 2017-11-15
Provide shellcode as a module function

jdevera:
As mentioned in #235 this can help with scripts that want to provide their own completion shell code to users.

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -8,6 +8,7 @@ from . import completers, my_shlex as shlex
 from .compat import USING_PYTHON2, str, sys_encoding, ensure_str, ensure_bytes
 from .completers import FilesCompleter, SuppressCompleter
 from .my_argparse import IntrospectiveArgumentParser, action_is_satisfied, action_is_open, action_is_greedy
+from .shellintegration import shellcode # noqa
 
 _DEBUG = "_ARC_DEBUG" in os.environ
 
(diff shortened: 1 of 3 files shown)


argcomplete/__init__.py
```diff
@@ -8,6 +8,7 @@
 from .compat import USING_PYTHON2, str, sys_encoding, ensure_str, ensure_bytes
 from .completers import FilesCompleter, SuppressCompleter
 from .my_argparse import IntrospectiveArgumentParser, action_is_satisfied, action_is_open, action_is_greedy
+from .shellintegration import shellcode
```
jdevera: The linter is complaining about this line, which I am using so that the `shellcode` functions is available at the topmost module level. Any suggestions?

argcomplete/__init__.py
```diff
@@ -8,6 +8,7 @@
 from .compat import USING_PYTHON2, str, sys_encoding, ensure_str, ensure_bytes
 from .completers import FilesCompleter, SuppressCompleter
 from .my_argparse import IntrospectiveArgumentParser, action_is_satisfied, action_is_open, action_is_greedy
+from .shellintegration import shellcode # noqa
```
jdevera: I have added noqa to avoid a linter error that is not actually an error. See https://bugs.launchpad.net/pyflakes/+bug/1178905

kislyuk:
LGTM but needs tests.

PR #241, opened 2017-12-05
Complete console scripts installed from wheels

### LINKED ISSUE #225: Global completion doesn't work for packages installed from wheels
[pytest](https://github.com/pytest-dev/pytest) has its command line scripts installed via [setuptools.setup(entry_points={'console_scripts': ...})](https://github.com/pytest-dev/pytest/blob/master/setup.py#L69) construction. If I understand correctly, argcomplete's global completion should handle this case and work just fine in such a situation, but it doesn't.
Looks like [this if statement](https://github.com/kislyuk/argcomplete/blob/master/argcomplete/bash_completion.d/python-argcomplete.sh#L49) should cover such a case, but `pytest` script created after installation doesn't have any of these strings in it. Here is the full content of `pytest` script from my environment:
```
#!/Users/pavel/.pyenv/versions/3.6.1/envs/test_env/bin/python3.6

# -*- coding: utf-8 -*-
import re
import sys

from pytest import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```
`pytest` module from which `main` function is imported has `# PYTHON_ARGCOMPLETE_OK` as its first string.

I've opened a ticket in pytest project (see https://github.com/pytest-dev/pytest/issues/2725), but it looks like that this is a problem of argcomplete.

evanunderscore:
Fixes #225 

Probably needs some tests, but I'm putting this up for discussion and so I don't forget about it.

### DIFF
--- a/argcomplete/_check_console_script.py
+++ b/argcomplete/_check_console_script.py
@@ -1,0 +1,42 @@
+import os
+import sys
+
+from pkg_resources import iter_entry_points
+
+from ._check_module import ArgcompleteMarkerNotFound, find
+
+
+def main():
+    # Argument is the full path to the console script.
+    script_path = sys.argv[1]
+
+    # Find the module and function names that correspond to this
+    # assuming it as actually a console script.
+    name = os.path.basename(script_path)
+    entry_points = list(iter_entry_points('console_scripts', name))
+    if not entry_points:
+        raise ArgcompleteMarkerNotFound('no entry point found matching script')
+    [entry_point] = entry_points
+    module_name = entry_point.module_name
+    function_name = entry_point.attrs[0]
+
+    # Check this looks like the script we really expected.
+    with open(script_path) as f:
+        script = f.read()
+    if 'from {} import {}'.format(module_name, function_name) not in script:
+        raise ArgcompleteMarkerNotFound('does not appear to be a console script')
+    if 'sys.exit({}())'.format(function_name) not in script:
+        raise ArgcompleteMarkerNotFound('does not appear to be a console script')
+
+    # Look for the argcomplete marker in the script it imports.
+    with open(find(module_name, return_package=True)) as f:
+        head = f.read(1024)
+    if 'PYTHON_ARGCOMPLETE_OK' not in head:
+        raise ArgcompleteMarkerNotFound('marker not found')
+
+
+if __name__ == '__main__':
+    try:
+        main()
+    except ArgcompleteMarkerNotFound as e:
+        sys.exit(e)
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,11 @@ matrix:
 
 before_install:
   - python -m pip install --quiet --upgrade pip codecov
+  # OSX tests require https://github.com/pypa/pip/pull/6008
+  # which is not in 19.1.1 (latest release at time of writing)
+  # in order to address https://stackoverflow.com/questions/24257803
+  # (the fix recommended there does not appear to work without it).
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then python -m pip install https://github.com/pypa/pip/archive/master.zip; fi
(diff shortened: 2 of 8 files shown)


mbdevpl:
@evanunderscore With latest release 1.10.0 the global completion for console scripts is still not working because of missing `# EASY-INSTALL-...` comment, as described in #225. And pip maintainers say in https://github.com/pypa/pip/issues/4702 that they will not do anything about it. So, I rebased this PR onto latest master (1.10.0), and now it's working! Thanks!

In my case I'm using pyenv.

I'm not sure why CI tests are failing, but in practice this PR definitely works, so it must be some corner case which doesn't occur in my environment. Is there any chance for the tests to be fixed/added and for merging this into master at some point?

evanunderscore:
The CI failure seems to be something unrelated on OSX, and codecov is complaining that I haven't tested enough. I'm short on free time, but I can try to get this up to scratch and merged soon.

evanunderscore:
@mbdevpl Can you verify this still works for you? I've made some very minor changes in the process of bringing this up to speed with master.

@kislyuk Would you like to review this? I don't believe we have any tests specifically targeting `python-argcomplete-check-easy-install-script` so I'd appreciate if you could sanity check that I haven't obviously broken that.

evanunderscore:
Unfortunately it looks like this OSX failure *is* related to the tests I've written, so I'll have to dig into that and see what the problem is.

evanunderscore:
@kislyuk Thanks for reviewing. I haven't managed to look at the failure yet but I'll report back as soon as I do.

evanunderscore:
@kislyuk I've managed to fix the issue, but it requires some fairly specific workarounds. I don't fully understand the problem but tried to link the related issues where relevant. If you're happy with this, it's good to go.

argcomplete/_check_console_script.py
```diff
@@ -0,0 +1,42 @@
...
+
+from ._check_module import ArgcompleteMarkerNotFound, find
+
+
+def main():
+    # Argument is the full path to the console script.
+    script_path = sys.argv[1]
+
+    # Find the module and function names that correspond to this
+    # assuming it as actually a console script.
```
kislyuk: "assuming it is"

argcomplete/_check_console_script.py
```diff
@@ -0,0 +1,42 @@
+import os
```
kislyuk: It would be nice to add a docstring here describing what this file does and linking to the definition and docs for console scripts.

evanunderscore:
Thanks for looking into this @kislyuk. I'll let you know when I've fixed up the review comments and pushed them to master.

> What happens if the user is running a Python that does not have pkg_resources available

Whoops, I'd assumed `pkg_resources` was built-in, but it looks like it's included with `setuptools`. If it's missing, the script will fail with an `ImportError` that will likely trigger automated error reporting we addressed in #261. However if we think this will only happen if the package is faulty, then that's probably for the best. Thoughts?

> What kind of worst-case real-life performance impact we can expect to see here, given that this script is going into the default code path for global completion

There are some minor performance implications related to the reordering of file accesses, but the important one is running the script. If tab completion is attempted on a Python script that is *not* meant to be tab completed via `argcomplete` one way or another, you'll end up down this branch, and you'll pay the performance cost of running the interpreter (and new script). However as touched on [here](https://github.com/kislyuk/argcomplete#common-problems), if you fail to complete any name, a default completer will get registered, so any subsequent attempts to tab complete that same name will have no penalty.

argcomplete/_check_console_script.py
evanunderscore: Addressed in #263

argcomplete/_check_console_script.py
evanunderscore: Addressed in #263

PR #245, opened 2018-01-15
document how to use argcomplete with zsh

evgeni:


evanunderscore:
Any objections to merging this? The CI failure is unrelated to the change, and I think it would be useful to have this documented in the README. Every time I troubleshoot zsh problems for somebody I have to dig up old issues where other people have described how they got it working. The only thing I'm not sure about is the `-U` flag; what does this do?

evanunderscore:
Thanks!

PR #246, opened 2018-01-25
Register completion for multiple commands

chrisdiamand:
Shell startup files may contain lots of 'register-python-argcomplete'
calls, for as many scripts as require it.

However, this can have a significant impact on shell startup time,
because each one requires starting a Python interpreter and invoking
'complete'.

Optimise this by allowing 'register-python-argcomplete' to accept
multiple commands. In Bash, the whole list of completable commands
can be passed to a single 'complete' call. Tcsh requires a separate
'complete' call for each command, but we can at least avoid the
overhead of starting Python each time.

On my machine, the time to register ~10 completed scripts decreases
from about 0.4s to 0.05s with this change.

### DIFF
--- a/argcomplete/shellintegration.py
+++ b/argcomplete/shellintegration.py
@@ -1,5 +1,10 @@
 #!/usr/bin/env python
 
+try:
+    from shlex import quote
+except ImportError:
+    from pipes import quote
+
 bashcode = r'''
 # Run something, muting output or redirecting it to the debug stream
 # depending on the value of _ARC_DEBUG.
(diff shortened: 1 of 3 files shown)


argcomplete/shellintegration.py
```diff
@@ -55,8 +55,12 @@ def shellcode(executable, use_defaults=True, shell='bash', complete_arguments=No
         complete_options = " ".join(complete_arguments)
 
     if shell == 'bash':
-        code = bashcode
+        quoted_executables = ['"%s"' % i for i in executables]
```
evanunderscore: Should use `shlex.quote` for this (it's small enough that you could copy and paste it for use on Python 2).

chrisdiamand:
Hi - thanks for the feedback. I think I've addressed that in the latest version, please let me know if I've misunderstood something.

Cheers!

kislyuk:
Sorry about the delay in merging this.

chrisdiamand:
Thanks!

PR #251, opened 2018-04-02
Get rid of non-standard streams (FD 8 and 9)

oOKexOo:
### Preface
First of all, thank you for this handy module.
However, I struggled a lot to get it working on Cygwin.
I traced back the issue to the non-standard stream (File descriptor 8), that cannot be opened:
https://github.com/kislyuk/argcomplete/blob/30ff207e9a89bc08b9cb38cbeb21147ef0c23480/argcomplete/__init__.py#L186-L191
### Changes
When this PR is meged it will:
- Get rid of the non-standard streams (FD 8 and 9).
- Handle the suppression of stderr and stdout in `__init__.py` rather than in the shell scripts.
### Additional Notes
- Additionally tested global completion with Bash on CYGWIN_NT-6.3.

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -12,7 +12,16 @@ from .shellintegration import shellcode
 
 _DEBUG = "_ARC_DEBUG" in os.environ
 
-debug_stream = sys.stderr
+debug_stream = sys.__stderr__
+
+# mute print function if _ARGCOMPLETE is defined and _DEBUG undefined
+# debug_stream and output_stream_default, however, are not muted
+if "_ARGCOMPLETE" in os.environ:
+    if _DEBUG:
+        sys.stdout = sys.stderr
+    else:
+        sys.stderr = None
+        sys.stdout = None
 
 def debug(*args):
     if _DEBUG:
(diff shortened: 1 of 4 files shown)


kislyuk:
Sorry. Cygwin is not supported. There is definitely an issue with using custom FDs (the client program might want to use them for something else), but the choice of custom FDs to redirect standard streams is very much intentional. There are many ways for the client program and its subprocesses to write to fds 1 and 2, and any such write that gets around any sort of filter that you install in-process in the middle of initialization (which is when argcomplete executes) will break argcomplete.

Using stdout/stderr to communicate is a no-go. It may be possible (but is unlikely) that using named pipes is a workable solution. In any case, if you're stuck on Windows, one possible solution is to switch from Cygwin to WSL.

oOKexOo:
Yeah, you are right. It is not guaranteed in any case that there is no print before the module is imported. Hence, my solution is deeply flawed.
On the bright side, I realized that the `/usr/bin/env python3` in my case was actually refering to the Windows build rather than the Cygwin one. When I switched to the Cygwin build, the original argcomplete works charming. Hence, only the Windows build cannot handle custom FDs. Named pipes would probably be the solution in that case. Unfortnatelly, the Windows Python build does not support `os.mkfifo` and probably requires the non-standard library `win32pipe` to do so.

Issue #68, opened 2013-12-30
Feature request: fish support


kwill:
Would it be possible to also support completions in fish? https://github.com/fish-shell/fish-shell

Lucretiel:
I will see if I can take a stab at this. As of at least 2.0, fish's `complete` method does support specifying a shell function as a custom completer (http://fishshell.com/docs/2.0/commands.html#complete)

wernight:
A simpler alternative may be to generate man page (see https://bugs.python.org/issue14102) and then ask Fish to complete from man page.

volkov:
Following code :
```fish
function __fish_myscript_complete
   set -x _ARGCOMPLETE 1
   set -x _ARGCOMPLETE_IFS \n
   set -x _ARGCOMPLETE_SUPPRESS_SPACE 1
   set -x COMP_LINE (commandline)
   set -x COMP_POINT (commandline -C) 
   myscript 8>&1 9>&2 1>&9 2>&1
end
complete -c myscript -f -a '(__fish_myscript_complete)'
```
in `~/.config/fish/completions/myscript.fish` works fine for me.

kislyuk:
Fish support has been released in v1.10.0, please test.

PR #260, opened 2019-01-05
Fish support

volkov:
Issue #68 work in progress.

Struck with problem of getting current process cursor position [superuser](https://superuser.com/questions/1390998/fish-commandline-get-cursor-position-in-current-process). 

Also some tests are broken (currently ignored or added to "expected failures").

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,12 @@ matrix:
     - language: generic
       python: 2.7
       os: osx
-      osx_image: xcode7.3
 
 before_install:
-  - pip install --quiet --upgrade pip codecov
+  - python -m pip install --quiet --upgrade pip codecov
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install bash; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install fish; fi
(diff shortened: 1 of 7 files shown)


volkov:
Well... Local run looks much better. I think I should check travis fish version/installation.

volkov:
@kislyuk Could you please take a look?

I also have to unlock xcode version (may be it should be locked on `xcode 9.4`?) and change `pip install` to `python -m pip install` to pass build.

evanunderscore:
This looks great! I'm not a fish user, but I'd be happy to take a look at this when I can find some spare time (though unfortunately that may not be for a while).

I can at least comment on the xcode problem - I added that line based on the advice in [this issue](https://github.com/travis-ci/travis-ci/issues/8829) because OSX builds were broken. If the builds are working again without it, it's fine to remove.

marcoesposito1988:
This would be a life saver.

Is any help needed?

kislyuk:
Sorry about the delay, didn't notice this was ready for review.

kislyuk:
I don't think we need to configure the OS X build beyond the defaults if it works.

kislyuk:
I'm waiting for tests to pass on master, then I will roll a release.

kislyuk:
Now released in v1.10.0. I'm not an active fish user, so please test this release for me.

@volkov thanks again for your work on this and apologies for the delays here.

PR #261, opened 2019-03-16
check_module: Don't crash, exit with error instead

hroncok:
This should avoid https://bugzilla.redhat.com/show_bug.cgi?id=1689507

I still don't know what makes this happen, or if a regular Exception is better here than AttributeError, but this is not the first time this was reported by Fedora automatic bug report.

See for example https://bugzilla.redhat.com/show_bug.cgi?id=1601763

I've asked the user what were they actually doing.

https://retrace.fedoraproject.org/faf/reports/2267286/ gives hint that this happens quite often.

### DIFF
--- a/argcomplete/_check_module.py
+++ b/argcomplete/_check_module.py
@@ -12,7 +12,10 @@ except ImportError:
 
     def find_spec(name):
         """Minimal implementation as required by `find`."""
-        f, path, _ = find_module(name)
+        try:
+            f, path, _ = find_module(name)
+        except ImportError:
+            return None
         has_location = path is not None
         if f is None:
             return ModuleSpec(None, has_location, [path])
@@ -20,17 +23,25 @@ except ImportError:
         return ModuleSpec(path, has_location, None)
 
 
+class ArgcompleteMarkerNotFound(RuntimeError):
+    pass
+
+
 def find(name):
     names = name.split('.')
     spec = find_spec(names[0])
+    if spec is None:
+        raise ArgcompleteMarkerNotFound(
+            'no module named "{}"'.format(names[0]))
     if not spec.has_location:
-        raise Exception('cannot locate file')
+        raise ArgcompleteMarkerNotFound('cannot locate file')
     if spec.submodule_search_locations is None:
         if len(names) != 1:
-            raise Exception('{} is not a package'.format(names[0]))
+            raise ArgcompleteMarkerNotFound(
+                '{} is not a package'.format(names[0]))
         return spec.origin
     if len(spec.submodule_search_locations) != 1:
-        raise Exception('expecting one search location')
+        raise ArgcompleteMarkerNotFound('expecting one search location')
     path = os.path.join(spec.submodule_search_locations[0], *names[1:])
     if os.path.isdir(path):
         return os.path.join(path, '__main__.py')
@@ -42,8 +53,11 @@ def main():
     with open(find(sys.argv[1])) as f:
         head = f.read(1024)
     if 'PYTHON_ARGCOMPLETE_OK' not in head:
-        raise Exception('marker not found')
+        raise ArgcompleteMarkerNotFound('marker not found')
 
 
 if __name__ == '__main__':
-    main()
+    try:
+        main()
+    except ArgcompleteMarkerNotFound as e:
+        sys.exit(e)


kislyuk:
@evanunderscore can you please take a look at this? Code in question was introduced in #204.

evanunderscore:
I had assumed that `find_spec` would raise an exception if it could not find the module as `find_module` does, but I missed the part in the documentation that says it returns `None`. This fix will at least make the cause of the failure clearer, so I'm all for it.

Having said that, failing with an exception is something this script is designed to do in any number of error cases. I don't know anything about Fedora's automatic bug reporting - what exactly is triggering that? How does it decide which kind of failures need to be reported?

Looking at the user's description, I can see why this is happening in his particular case, and I can probably come up with a patch that checks how many arguments have been passed before trying to invoke some of these completion methods, but there are still other expected failure cases that will likely trigger this reporting mechanism, so I'd like to get to the bottom of that if possible. Can you provide any insight @hroncok?

hroncok:
The Fedora's automatic bug reporting recognizes if a program crashes with a traceback on stderr and non-0 exit code. It usually means a problem. If exiting with non-0 is desired here, to avoid the report, the code would need to catch the exception and report the error via print to stdout or similar. I can send a PR, but I like to know: how is this failure detected elsewhere?

hroncok:
Something like this:

* convert all risen Exceptions to something more specific (RuntimeError amybe?)

```python
try:
    main()
except RuntimeError as e:
    print(e, file=sys.stderr)
    sys.exit(1)
```

evanunderscore:
> how is this failure detected elsewhere?

This is an internal function that is run by the global completion function to check whether a given module is compatible with argcomplete. You can see where it is invoked here:

https://github.com/kislyuk/argcomplete/blob/5e4d3c6a05885c9085b8ca96d6e3ef37b3db638b/argcomplete/bash_completion.d/python-argcomplete.sh#L32

If it's the only way to avoid the automatic bug report, then yes, we'd have to catch the exceptions and explicitly exit with a non-zero error code instead. In your snipped above, `sys.exit(e)` will handle the printing part for you and exit with code 1. Rather than `RuntimeError` I'd prefer to define our own exception type and specifically throw/catch that to avoid catching something we don't mean to. The find_spec reimplementation would also need to handle the `ImportError` potentially raised by `find_module` and instead return `None`.

Would you be happy to implement that?

hroncok:
I would! Thanks.

hroncok:
Done.

argcomplete/_check_module.py
```diff
@@ -12,25 +12,32 @@
...
+        except ImportError:
+            return None
         has_location = path is not None
         if f is None:
             return ModuleSpec(None, has_location, [path])
         f.close()
         return ModuleSpec(path, has_location, None)
 
 
+class LocateFailure(RuntimeError):
```
hroncok: Better name suggestions appreciated.

argcomplete/_check_module.py
```diff
@@ -12,25 +12,32 @@
...
+    pass
+
+
 def find(name):
     names = name.split('.')
     spec = find_spec(names[0])
-    if not spec.has_location:
-        raise Exception('cannot locate file')
+    if spec is None or not spec.has_location:
+        raise LocateFailure('cannot locate file')
```
evanunderscore: It would make for nicer debugging if we split out these two cases - `None` means no module with that name was found, whereas `has_location` being `False` means a module was found but it does not have a corresponding file on disk.

argcomplete/_check_module.py
evanunderscore: Perhaps `CheckFailure` or `MarkerNotFound` given that it's a generic exception to indicate that *something* has gone wrong during the process of checking the given module name to see if it contains the PYTHON_ARGCOMPLETE_OK marker. Don't think I have a strong preference for any of these three though.

argcomplete/_check_module.py
```diff
@@ -12,25 +12,34 @@
...
 
+class LocateFailure(RuntimeError):
+    pass
+
+
 def find(name):
     names = name.split('.')
     spec = find_spec(names[0])
+    if spec is None:
+        raise LocateFailure('no module named "{}"'.format(name))
```
evanunderscore: Nitpicking, but the thing we failed to find is `names[0]` (as opposed to `name`)

evanunderscore:
@hroncok Thanks for your work on this, the changes look good to me. Up to you whether you want to do anything about the remaining unresolved comments. Let me know when you're ready for me to merge.

hroncok:
I will address the comments.

argcomplete/_check_module.py
hroncok: I like `MarkerNotFound`. I also thought about  `CannotArgcomplete` or even `NoArgcomplete`.

evanunderscore: They're all fine by me. I'm happy for @kislyuk to weigh in with a preference, or you can take your pick.

kislyuk: I would go with `ArgcompleteMarkerNotFound`.

kislyuk:
@hroncok Please update with the preferred wording for the error name. Otherwise, is this good to merge?

kislyuk:
(Sorry about the delay in my response.)

argcomplete/_check_module.py
hroncok: Amended.

kislyuk:
Released in v1.9.5, please test.

hroncok:
I'll update the Fedora package and hopefully those reports will stop.

PR #264, opened 2019-08-08
Help string support for fish

volkov:
Help string support for fish.

Work in progress.

`dfs` and `_ARGCOMPLETE_DFS` should probably be renamed to something more meaningful.

2019-08-08-162057_575x69_scrot


argcomplete/__init__.py
```diff
@@ -222,6 +227,12 @@ def __call__(self, argument_parser, always_complete_options=True, exit_method=os
 
         completions = self._get_completions(comp_words, cword_prefix, cword_prequote, last_wordbreak_pos)
 
+        if dfs:
+            display_completions = {key_part: value
+                                   for key, value in self.get_display_completions().items()
+                                   for key_part in key.split(" ")}
```
volkov: I think it's possible to have options with spaces, so there would be no help string for them.
Not sure if it is reasonable to fix because it would take much more changes - I have to change `get_display_completions()` method or add very similar method (that would not join options in [_get_option_completions](https://github.com/kislyuk/argcomplete/blob/master/argcomplete/__init__.py#L348) )

kislyuk:
@volkov thanks for your work on this, and sorry about the delay in my review.

This looks good overall. Can you add a description of this feature, and a rendering of what it looks like in the shell, in the project README under the "fish support" section?

Also, it's certainly possible that completions will contain spaces (I'm sure you realize that completions can be produced for values, not just option names). In those cases, can you describe and paste renderings of what happens? How does this code interact with completions that contain both quotes and spaces? Since the only completions that contain spaces are value completions, and those are unlikely to have help, perhaps it's better to skip them instead of splitting them the way you are doing?

volkov:
@kislyuk thanks for your review

> Can you add a description of this feature, and a rendering of what it looks like in the shell, in the project README under the "fish support" section?

Yes, before finishing work on it.

> Also, it's certainly possible that completions will contain spaces (I'm sure you realize that completions can be produced for values, not just option names). In those cases, can you describe and paste renderings of what happens? How does this code interact with completions that contain both quotes and spaces? Since the only completions that contain spaces are value completions, and those are unlikely to have help, perhaps it's better to skip them instead of splitting them the way you are doing?

I can't find how to set help string for value completions. Tried to use `dict` as value for choices, but it does not help. There is no clue in [documentation](https://docs.python.org/3/library/argparse.html#choices) also. Do you know how to set it?

I have added space to subparser name:
```python
subparsers.add_parser('basic with space', help="basic help").add_argument('arg', choices=['foo', 'bar', 'baz'])
```
In this it drops help message:
2019-08-26-072626_569x73_scrot
Same with quotes:
2019-08-26-073431_1524x109_scrot




I found another issue - help message could contain `_ARGCOMPLETE_DFS` or `_ARGCOMPLETE_IFS` - in this case additional completions are added.

volkov:
So I fixed problem with ifs in descriptions and added some documentation.

But I still don't know good solution space handling.

If I don't split I miss description for options with aliases, e.g help here:
2020-01-14-232143_1158x158_scrot

If I split (this is currently realized in PR) I miss descriptions for subcommands with spaces
2020-01-14-232024_1415x153_scrot

evanunderscore:
From my reading, it looks like the original is trying to display all of the aliases together, e.g. `-h --help -- show this help and exit`, whereas you're trying to duplicate the help message for each parameter, e.g. `-h (show this help message and exit)` and `--help (show this help message and exit)`.

Since the previous API hides the internal dict behind a function call `get_display_completions`, nothing should stop you from modifying what's stored in `self._display_completions` and then adding whatever logic you need in that function to restore it to what it was previously.

For example (not tested):

```
    def _get_option_completions(self, parser, cword_prefix):
        self._display_completions.update(
            # replace " ".join with tuple
            [[tuple(ensure_str(x) for x in action.option_strings
                       if ensure_str(x).startswith(cword_prefix)), action.help]
             for action in parser._actions
             if action.option_strings])
```


```
def get_display_completions():
    # " ".join the tuple to preserve previous API
    return {" ".join(k): v for k, v in self._display_completions}
```

```
        if dfs:
            display_completions = {key_part: value.replace(ifs, " ") if value else ""
                                   # use self._display_completions directly
                                   for key, value in self._display_completions.items()
                                   # no need to split key since it is already a tuple
                                   for key_part in key}
```

The actual change will be a little more involved since you'll either have to find all the other places in the code adding string keys and turn them into one element tuples, or account for both strings and tuples when you use the dict.

volkov:
@evanunderscore that's good idea, I had changed pull request accordingly.

volkov:
And now I'm investigating problem with my code around `argcomplete.CompletionFinder._get_subparser_completions`

argcomplete/__init__.py
```diff
@@ -321,18 +332,10 @@ def __call__(self, parser, namespace, values, option_string=None):
...
-            a = metavar.replace(",", "").split()
-            return " ".join(x for x in a if x.startswith(prefix))
-
         for action in parser._get_subactions():
-            subcmd_with_aliases = filter_aliases(action.metavar, action.dest, cword_prefix)
-            if subcmd_with_aliases:
-                self._display_completions[subcmd_with_aliases] = action.help
+            dest = action.dest
+            if dest and dest.startswith(cword_prefix):
+                self._display_completions[dest] = action.help
```
evanunderscore: Should this be `(dest,)`?

argcomplete/__init__.py
```diff
@@ -321,18 +332,10 @@ def __call__(self, parser, namespace, values, option_string=None):
         return self.active_parsers
 
     def _get_subparser_completions(self, parser, cword_prefix):
-        def filter_aliases(metavar, dest, prefix):
-            if not metavar:
-                return dest if dest and dest.startswith(prefix) else ""
-
-            # metavar combines dest and aliases with ",".
-            a = metavar.replace(",", "").split()
-            return " ".join(x for x in a if x.startswith(prefix))
```
evanunderscore: This was handling subparser aliases. Example from https://docs.python.org/3/library/argparse.html:

```
>>> parser = argparse.ArgumentParser()
>>> subparsers = parser.add_subparsers()
>>> checkout = subparsers.add_parser('checkout', aliases=['co'])
>>> checkout.add_argument('foo')
>>> parser.parse_args(['co', 'bar'])
Namespace(foo='bar')
```

It stores the display for these in the metavar attribute as e.g. "checkout (co)". If there's more than one, it joins them with a comma, e.g. "checkout (co, other)".

The removed code doesn't actually handle this correctly because it doesn't account for the braces. A more robust solution might be to inspect `parser._name_parser_map` to see which names map to the same parser as `parser._name_parser_map[dest]`. If we want to maintain strict backwards compatibility, we'd have to also store the existing display text alongside the tuple, otherwise we can't guarantee we'd reconstruct it in the correct order. Alternatively if we're happy to accept a minor break, you could just join the names back together as you are with the others.

volkov: Thank you for explanation, covered current behaviour in https://github.com/kislyuk/argcomplete/pull/295/files

volkov: Added ugly fix `a = metavar.replace(" (", ", ").replace(")", "").split(", ")` which guarantees order but still break compatibility because of missing braces.

volkov: I rewritten code using `parser.choices`. @evanunderscore take a look, please

argcomplete/__init__.py
```diff
@@ -321,16 +332,19 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+        def filter_aliases(aliases, prefix):
+            return tuple(x for x in aliases if x.startswith(prefix))
+
+        aliases_by_parser = {}
+        for key in parser.choices.keys():
+            p = parser.choices[key]
+            if not aliases_by_parser.get(p):
```
volkov, line 341: Probably should change this to `defaultdict`

argcomplete/__init__.py
evanunderscore: Issue has been fixed and a test case added. Thanks!

argcomplete/__init__.py
evanunderscore: Agree, or you could use `aliases_by_parser.setdefault(p, []).append(key)`

argcomplete/__init__.py
evanunderscore: Looks good to me, thanks!

evanunderscore: I've merged your other PR with the test case. Can you merge/rebase this branch so that the change is more obvious in the diff please?

volkov: done

argcomplete/__init__.py
volkov: changed to `setdefault`

evanunderscore:
This looks great! Thanks for your persistence with this @volkov. I'm happy with the alias changes as illustrated by the changes to the test code, but since this is technically a breaking change I'll leave it to @kislyuk to sign off on it.

volkov:
@kislyuk have a look please

kislyuk:
LGTM - thanks for sticking with this PR through the long development cycle @volkov. I'll roll a release shortly.

PR #266, opened 2019-08-22
Skip tests for fish and tcsh when they are not installed

kapsh:


kislyuk:
Thank you for your PR. It will not be accepted in its current form because silently skipping the tests like this would compromise the ability of the test suite to catch regressions. You can run a subset of the tests by passing class names to the unittest test runner. I would also be open to a conditional like `@unittests.skipUnless(json.loads(os.envirion.get("ARGCOMPLETE_SKIP_TCSH_TESTS")))`, which would require a variable to positively be set to skip the tests, instead of an executable to be absent.

PR #269, opened 2019-10-02
Catch exceptions in _check_module

vstinner:
* _check_module.main() now catchs IndexError and FileNotFoundError
  errors: raise ArgcompleteMarkerNotFound exception instead.
* Replace open() with tokenize.open() to handle "# coding: xxx"
  encoding cookie: support source code using an encoding different
  than UTF-8 on Python 3.

### DIFF
--- a/argcomplete/_check_module.py
+++ b/argcomplete/_check_module.py
@@ -68,10 +68,15 @@ def main():
         raise ArgcompleteMarkerNotFound('missing argument on the command line')
 
     filename = find(name)
+    if hasattr(tokenize, 'open'):
+        open_func = tokenize.open
+    else:
+        open_func = open
+
     try:
-        fp = tokenize.open(filename)
-    except FileNotFoundError:
-        raise ArgcompleteMarkerNotFound('file not found')
+        fp = open_func(filename)
+    except OSError:
+        raise ArgcompleteMarkerNotFound('cannot open file')
 
     with fp:
         head = fp.read(1024)
--- a/argcomplete/_check_module.py
+++ b/argcomplete/_check_module.py
@@ -8,6 +8,7 @@ Intended to be invoked by argcomplete's global completion function.
 """
 import os
 import sys
+import tokenize
 
 try:
     from importlib.util import find_spec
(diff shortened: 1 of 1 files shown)


vstinner:
On Fedora, a default exception hook is installed on the whole system. Uncatched exceptions are reported to Fedora using ABRT. Type "python3 -m http.ser[TAB]" on Fedora detects a bug and invites the user to report the bug to Fedora. Example of such report:
https://bugzilla.redhat.com/show_bug.cgi?id=1737212

Bonus: this change also adds support for encoding different than UTF-8 on Python 3. Tell me if you prefer a separated PR for that.

Note: I tested my change on Python 3.7 using my other PR #268.

vstinner:
I rebased this PR on top on my tox change.

Oops, argcomplete supports Python 2: I didn't notce that. I adapted my PR for Python 2.

argcomplete/_check_module.py
```diff
@@ -61,8 +62,25 @@ def find(name, return_package=False):
 
 
 def main():
-    with open(find(sys.argv[1])) as f:
-        head = f.read(1024)
+    try:
+        name = sys.argv[1]
+    except IndexError:
+        raise ArgcompleteMarkerNotFound('missing argument on the command line')
```
evanunderscore: This is only ever called [here](https://github.com/kislyuk/argcomplete/blob/b05ec4068830c9654bf5758bf5156ab78655df46/argcomplete/bash_completion.d/python-argcomplete.sh#L32) and it always passes an argument, even if that argument is the empty string.

I'm unfamiliar with Fedora's bug reporting; will it trigger even if the user types `python -m argcomplete._check_module` at the prompt? If so, this change is a reasonable protection against that. If not, I don't think it's necessary.

evanunderscore:
Thanks for the PR @vstinner! I'm happy to merge this for you, however I won't be able to push a release.

argcomplete/_check_module.py
vstinner: https://bugzilla.redhat.com/show_bug.cgi?id=1737212 bug report was only about unhandled open() error. You can modify my PR to remove error checking on sys.argv if you prefer.

vstinner:
I forgot this PR that I wrote 2 months ago. Feel free to modify it if you want :-)

kislyuk:
@evanunderscore your discretion with this PR, please merge as you see fit and I will cut a release when ready.

argcomplete/_check_module.py
evanunderscore: On reflection, I'm okay with leaving this in. If this were using argparse, it would end up calling `sys.exit(1)` anyway, so this has roughly the same effect.

evanunderscore:
Thanks for your work @vstinner!

PR #270, opened 2019-10-16
Disable option completion after seeing --

dcolascione:


kislyuk:
Thanks, this seems like a legitimately needed PR but it will need lint fixes and a test case in order to proceed.

PR #275, opened 2019-11-21
Prepend to PYTHONPATH in test_console_script* instead of replacing

mgorny:
Modify _test_console_script() to prepend to PYTHONPATH instead of
overwriting its value.  This is necessary so that it is able to find
just-built version of argcomplete that is provided via PYTHONPATH
(vs requiring it to be installed in system-wide directory prior).

Otherwise, the tests are failing:

      File "./bin/test-package", line 5, in <module>
        from test_package import main
      File "/tmp/argcomplete/test_dir_pychyg9vqx/test_package/__init__.py", line 4, in <module>
        import argcomplete
    ModuleNotFoundError: No module named 'argcomplete'


kislyuk:
Do you require me to cut a release?

mgorny:
No need to, already patched it on our end.

PR #279, opened 2019-12-19
Make arcomplete work with Spack-created console scripts

mbdevpl:
Spack (https://spack.io/) is a package manager for supercomputers.
To avoid overlong shebang problem in some environments
it relies on a trick with wrapping exec in a Python docstring.

Python console scripts have the following form when created by Spack:

```
#!/bin/sh
'''exec' /very/long/path/to/python/installed/within/spack/bin/python3.8 "$0" "$@"
' '''
# -*- coding: utf-8 -*-
import re
import sys
from my_package.__main__ import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

I'm not sure if this PR is general enough to be merged, but I hope that some other users would find it useful.

### DIFF
--- a/argcomplete/bash_completion.d/python-argcomplete.sh
+++ b/argcomplete/bash_completion.d/python-argcomplete.sh
@@ -54,6 +54,9 @@ _python_argcomplete_global() {
             elif __python_argcomplete_run "$interpreter" -m argcomplete._check_console_script "$SCRIPT_NAME"; then
                 local ARGCOMPLETE=1
             fi
+        elif [[ "$(head -n 1 "$SCRIPT_NAME")" =~ ^"#"!/bin/sh$ ]] && [[ "$(head -n 2 "$SCRIPT_NAME" | tail -n 1)" =~ ^\'\'\'exec\'.*(python|pypy)[0-9\.]*.*$ ]]; then
+            # Spack alters shebang using a trick with exec
+            local ARGCOMPLETE=1
         fi
     fi
 


kislyuk:
I'm sorry but I don't see this being an appropriate patch in its current form.

The global completion hook looks for a positive signal that the target executable supports argcomplete. This is necessary because in search for completions, we **execute the target** - a potentially destructive operation. As a compromise, we have accommodated several (probably misguided) tools that hide the actual executable behind layers of wrappers. We traverse these wrappers to their target only when we are certain that the script is a wrapper and that we can collect enough information on the other side. To be clear, my inclination is not to support those tools at all. Modern pip/setuptools installs scripts that contain the actual payload, including our magic string. The only tool that is widely used that we need to traverse wrappers for is setuptools entry_points, and we can do that with reasonable certainty thanks to the logic in https://github.com/kislyuk/argcomplete/blob/master/argcomplete/_check_console_script.py and the package metadata that it is able to query.

In your case, the signal that you're looking for to see if the wrapper script should be followed is just not specific enough. The presence of `!/bin/sh` in the first line and `exec python` in the first two lines is not enough signal. To execute a target based on just that information in search for completions would be reckless. What if it's a shell script of some sort instead, and executing it is harmful to the system?

The workaround is to register your scripts explicitly. Your issue is only related to global completion; individual registration with `register-python-argcomplete PATH` should still work. Please open a separate issue if that is not the case.

PR #280, opened 2019-12-20
Use shell builtins where possible

### LINKED ISSUE #278: bash: warning: command substitution: ignored null byte in input
After installing argcomplete I get the following warning from bash frequently. Everything still works but this is very annoying. It happens the instant I hit tab to auto-complete and it happens about 10% of the time. Any idea what's causing this? 

```
bash: warning: command substitution: ignored null byte in input
```

kislyuk:
Fixes #278

### DIFF
--- a/argcomplete/bash_completion.d/python-argcomplete
+++ b/argcomplete/bash_completion.d/python-argcomplete
@@ -22,6 +22,15 @@ __python_argcomplete_run() {
     fi
 }
 
+# Scan the beginning of an executable file ($1) for a regexp ($2). By default,
+# scan for the magic string indicating that the executable supports the
+# argcomplete completion protocol. By default, scan the first kilobyte;
+# if $3 is set to -n, scan until the first line break up to a kilobyte.
+__python_argcomplete_scan_head() {
+    local HEAD=$((read -s -r ${3:--N} 1024 < "$1"; echo "$REPLY") 2>/dev/null)
+    [[ "$HEAD" =~ ${2:-PYTHON_ARGCOMPLETE_OK} ]]
+}
+
 _python_argcomplete_global() {
     local executable=$1
     __python_argcomplete_expand_tilde_by_ref executable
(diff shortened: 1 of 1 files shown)


argcomplete/bash_completion.d/python-argcomplete.sh
```diff
@@ -34,21 +47,21 @@ _python_argcomplete_global() {
...
-        elif [[ -f "${COMP_WORDS[1]}" ]] && (head -c 1024 "${COMP_WORDS[1]}" | grep --quiet "PYTHON_ARGCOMPLETE_OK") >/dev/null 2>&1; then
+        elif [[ -f "${COMP_WORDS[1]}" ]] && __python_argcomplete_scan_head_noerr "${COMP_WORDS[1]}"; then
             local ARGCOMPLETE=2
         else
             return
         fi
-    elif which "$executable" >/dev/null 2>&1; then
-        local SCRIPT_NAME=$(which "$executable")
+    elif type -P "$executable" >/dev/null 2>&1; then
+        local SCRIPT_NAME=$(type -P "$executable")
```
evanunderscore: I don't think this is quite right now - `type -P` will find a file even if shadowed by something else, causing the below block to search that file for the marker, but if it finds it, it will not actually be able to run it. Unfortunately `type -p "$executable"` appears to exit with code 0 if the thing exists but is not a file, so you'd have to check if its output is not blank, or alternatively check if `type -t "$executable"` is "file".

argcomplete/bash_completion.d/python-argcomplete.sh
```diff
@@ -34,21 +47,21 @@ _python_argcomplete_global() {
...
         if (type -t pyenv && [[ "$SCRIPT_NAME" = $(pyenv root)/shims/* ]]) >/dev/null 2>&1; then
             local SCRIPT_NAME=$(pyenv which "$executable")
         fi
-        if (head -c 1024 "$SCRIPT_NAME" | grep --quiet "PYTHON_ARGCOMPLETE_OK") >/dev/null 2>&1; then
+        if __python_argcomplete_scan_head_noerr "$SCRIPT_NAME"; then
             local ARGCOMPLETE=1
-        elif [[ "$(head -n 1 "$SCRIPT_NAME")" =~ ^#!(.*)$ ]] && [[ "${BASH_REMATCH[1]}" =~ ^.*(python|pypy)[0-9\.]*$ ]]; then
+        elif __python_argcomplete_scan_head_noerr "$SCRIPT_NAME" "^#!(.*)$" -n && [[ "${BASH_REMATCH[1]}" =~ ^.*(python|pypy)[0-9\.]*$ ]]; then
```
evanunderscore: Consider using single quotes where you don't explicitly want variable expansion etc. This regular expression works within the script, but would confuse somebody trying to replicate it in an interactive shell as the ! will cause history expansion.

argcomplete/bash_completion.d/python-argcomplete.sh
```diff
@@ -34,21 +47,21 @@ _python_argcomplete_global() {
...
             local interpreter="$BASH_REMATCH"
-            if (head -c 1024 "$SCRIPT_NAME" | egrep --quiet "(PBR Generated)|(EASY-INSTALL-(SCRIPT|ENTRY-SCRIPT|DEV-SCRIPT))" \
+            if (__python_argcomplete_scan_head_noerr "$SCRIPT_NAME" "(PBR Generated)|(EASY-INSTALL-(SCRIPT|ENTRY-SCRIPT|DEV-SCRIPT))" \
                 && "$interpreter" "$(which python-argcomplete-check-easy-install-script)" "$SCRIPT_NAME") >/dev/null 2>&1; then
```
evanunderscore: Option to also replace this `which` with `type -p`.

argcomplete/bash_completion.d/python-argcomplete.sh
kislyuk: I'm not sure if I follow. Can you give a specific situation where `type -P` behaves differently from `which`? Because to me it seems that `which` is just as susceptible to the problem you describe.

argcomplete/bash_completion.d/python-argcomplete.sh
kislyuk: Changed to single quotes.

argcomplete/bash_completion.d/python-argcomplete.sh
kislyuk: Replaced `which` with `type -P` (not `type -p` - as per my question above, I'm not sure why we would use `type -p` in place of `which`).

kislyuk:
@evanunderscore I'm going to merge this, but I do want to know what you think about my question regarding `type -P` vs. `type -p` above. I'm thinking `type -P` behaves more like `which` than any other built-in alternative, but if we find a scenario where that's not the case, I'm totally open to replacing it with a more compatible multi-line construct.

argcomplete/bash_completion.d/python-argcomplete.sh
evanunderscore: My mistake - you're right that they both suffer from the same problem. In which case what I'm suggesting is actually an improvement over what we had before, since `type -p` is able to tell whether the thing you're trying to execute is actually a file.

As an example, suppose you create an alias which shadows a script on your path that is able to be completed by argcomplete. If the user types `myalias <tab>`, completion will complete the script, however when the user actually tries to execute it, it will execute the alias.

(As an aside, I've just noticed that currently we're wrapping `$executable` in quotes when we execute it, which is actually incorrect, and means we won't correctly operate on things like aliases and things with quotes and escaped characters. If we remove that, we'll actually execute the alias during tab completion, which will be even worse.)

evanunderscore: I dug into this a little further and realised it's not really to do with whether `$executable` is quoted or not. We'd need to do something like [this](https://stackoverflow.com/questions/6988288/linux-bash-how-to-unquote) if we want aliases and special characters to work.

Issue #282, opened 2019-12-20
Speed up completion hook shellcode


kislyuk:
@evanunderscore interest

In the course of looking into #278, I've become concerned about the performance impact of `pkg_resources.iter_entry_points()` in https://github.com/kislyuk/argcomplete/blob/master/argcomplete/_check_console_script.py.

I'd like us to profile the completion hook shellcode. We might want to use techniques developed by 
https://github.com/DropD/reentry to speed up entry point scanning, or look for other ways to speed up the shellcode.

kislyuk:
(Related: https://github.com/pypa/setuptools/issues/510)

evanunderscore:
For what it's worth, the penalty will only be paid if the thing you're completing is a file with a shebang line for Python. In the case that we don't actually want to complete it, the price is paid once per name per shell, after which point a default completer is registered and the name will not be checked again.

I went with the `entry_points` solution based on the discussion on https://github.com/pypa/pip/issues/4702. If there were an easier way to tell what a console entry point script looked like ahead of time, we could avoid paying that price in most cases.

It sounds like `importlib_metadata` will improve performance. Worth noting is that it seems to support 2.7 and 3.5+, however argcomplete still claims support for 3.3+ in the README. We could fall back on the `pkg_resources` method for 3.3/3.4, or could drop this particular feature for these versions.

Supposing we continue to use `argcomplete._check_module`, note that everything in `argcomplete.__init__` will always be unnecessarily executed in the process of importing it. It's worth profiling and seeing how much of a performance penalty that imposes too. If it's significant, we could avoid it by either moving `argcomplete._check_module` to a separate script like `python-argcomplete-check-easy-install-script`, or by moving most of `__init__` into a separate submodule.

For scripts we do actually want to complete, we are currently running this script logic every single time. We do have the option to register a separate completer that effectively caches the decision to complete and how, e.g. by running `register-python-argcomplete` which will avoid having to run the global completion script again for that name for the lifetime of the shell. The downside of this is that if the underlying script is removed, or changes to one that is not intended to be completed, that decision will still be cached and it will attempt to execute it anyway.

Also worth considering, we could make this script completion feature opt-in/out with a flag to `activate-global-python-argcomplete`.

evanunderscore:
I forgot I'd actually prototyped the decision caching at some point. You can see the changes here: https://github.com/evanunderscore/argcomplete/tree/cachedecision. Note this won't work for module completion but it would work for entry points.

[...]

PR #283, opened 2019-12-20
Switch from pkg_resources to importlib

kislyuk:
On my workstation, this produces a 40% speedup.

Related to #282

### DIFF
--- a/argcomplete/_check_console_script.py
+++ b/argcomplete/_check_console_script.py
@@ -14,7 +14,10 @@ Intended to be invoked by argcomplete's global completion function.
 import os
 import sys
 
-from pkg_resources import iter_entry_points
+try:
+    from importlib.metadata import entry_points as importlib_entry_points
+except ImportError:
+    from importlib_metadata import entry_points as importlib_entry_points
 
 from ._check_module import ArgcompleteMarkerNotFound, find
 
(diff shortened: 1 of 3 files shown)


argcomplete/_check_console_script.py
```diff
@@ -26,12 +29,11 @@ def main():
...
     # assuming it is actually a console script.
     name = os.path.basename(script_path)
-    entry_points = list(iter_entry_points('console_scripts', name))
+    entry_points = [ep for ep in importlib_entry_points()["console_scripts"] if ep.name == name]
     if not entry_points:
         raise ArgcompleteMarkerNotFound('no entry point found matching script')
-    [entry_point] = entry_points
-    module_name = entry_point.module_name
-    function_name = entry_point.attrs[0]
+    entry_point = entry_points[0]
```
evanunderscore: I originally wrote this to fail if there are two entry points for the same name found, since I don't know what actually happens in that case. I suspect one will overwrite the other, and then we don't really have a way of knowing which one to check for the marker. Perhaps we should throw `ArgcompleteMarkerNotFound` if there's more than one?

setup.py
```diff
@@ -35,10 +42,10 @@
         'Programming Language :: Python :: 2',
         'Programming Language :: Python :: 2.7',
         'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.3',
-        'Programming Language :: Python :: 3.4',
```
evanunderscore: Should also update the README to say we support 3.5+ (currently says 3.3+).

argcomplete/_check_console_script.py
kislyuk: On my computer, I get multiple entry points for packages that I have installed multiple times. I'm not sure what causes that, but it doesn't seem to me that we need to treat it as an error.

setup.py
kislyuk: Done.

kislyuk:
Despite the codecov report, this code does get run by the test suite, just in a subprocess so the coverage doesn't seem to be reported.

argcomplete/_check_console_script.py
evanunderscore: There's nothing that stops multiple packages specifying the same `entry_point` name, in which case you could be checking for the marker in one script but executing another.

Issue #67, opened 2013-12-27
~/.bash_completion.d/python-argcomplete.sh prevents other completions


jmlopez-rod:
I noticed a while ago that when I source the script `python-argcomplete.sh` I cannot tab to complete environment variables. If I comment the last line in the script:

```
#complete -o default -D -F _python_argcomplete_global
```

Then I can do:

```
macbook-pro:~ jmlopez$ echo $PA<TAB>
macbook-pro:~ jmlopez$ echo $PATH
```

As soon as I activate the global completer then it doesn't do the environment variable completion anymore. This behavior does not happen when registering an awesome script, i.e.

```
eval "$(register-python-argcomplete my-awesome-script.py)"
```

Not sure if this is happening with others, can someone check if this is  bug on argparse? I'm using bash 4.2:

```
macbook-pro:~ jmlopez$ echo $BASH_VERSION
4.2.45(2)-release
```

jiasli:
`register-python-argcomplete` is not using `-o bashdefault` by default:

https://github.com/kislyuk/argcomplete/blob/abfa18712ec795445857f4125ed5ff3b51107f2c/argcomplete/shell_integration.py#L77

Although it can be overridden with [`--complete-arguments`](https://github.com/kislyuk/argcomplete/blob/b2a3192d6804986a4b51a0e355d5bb0cc1f7c6eb/scripts/register-python-argcomplete#L38-L41), maybe we can include `-o bashdefault` for individual command registration as well?

evanunderscore:
Thank you for reporting @jiasli. This appears to be an oversight. Would you like to submit a PR?

jiasli:
Sure. With pleasure!

PR #284, opened 2020-01-10
Add -o bashdefault to register-python-argcomplete's output command

jiasli:
Fix https://github.com/kislyuk/argcomplete/issues/67#issuecomment-570475793

Add [`-o bashdefault`](https://www.gnu.org/software/bash/manual/html_node/Programmable-Completion-Builtins.html) to `register-python-argcomplete`'s output command, so that individual command can complete the environment variables.

### DIFF
--- a/argcomplete/shell_integration.py
+++ b/argcomplete/shell_integration.py
@@ -74,7 +74,7 @@ def shellcode(executables, use_defaults=True, shell='bash', complete_arguments=N
     '''
 
     if complete_arguments is None:
-        complete_options = '-o nospace -o default -o bashdefault' if use_defaults else '-o nospace'
+        complete_options = '-o nospace -o default -o bashdefault' if use_defaults else '-o nospace -o bashdefault'
     else:
         complete_options = " ".join(complete_arguments)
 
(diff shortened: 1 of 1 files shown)


argcomplete/shell_integration.py
```diff
@@ -74,7 +74,7 @@ def shellcode(executables, use_defaults=True, shell='bash', complete_arguments=N
     '''
 
     if complete_arguments is None:
-        complete_options = '-o nospace -o default' if use_defaults else '-o nospace'
+        complete_options = '-o nospace -o default -o bashdefault' if use_defaults else '-o nospace'
```
evanunderscore: We should also include `-o bashdefault` when `use_defaults` is False to match the behaviour of the global completion script.

jiasli: Nice catch. Added as requested.

evanunderscore:
Perfect, thanks!

kislyuk:
Released in v3.3.3, please test.

kislyuk:
Released in v1.11.1, please test.

PR #288, opened 2020-02-11
Register auto completion for an arbitrary name using a given external script

msc42:
Allows to use the auto completion functionality of argcomplete for an application not written in Python. The command line interface of this program must be additionally implemented in a script with argparse and argcomplete and whenever the application is called the registered external script is used for auto completion.

registration:
eval "$(register-python-argcomplete --external-argcomplete-script /path/to/external_script.py arbitrary_app)"

### DIFF
--- a/argcomplete/shell_integration.py
+++ b/argcomplete/shell_integration.py
@@ -91,7 +91,7 @@ def shellcode(executables, use_defaults=True, shell='bash', complete_arguments=N
         code = ""
         for executable in executables:
             if not argcomplete_script:
-               argcomplete_script = executable
+                argcomplete_script = executable
             code += shell_codes.get(shell, '') % dict(executable=executable, argcomplete_script=argcomplete_script)
 
     return code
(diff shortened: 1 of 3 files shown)


evanunderscore:
The test failures appear to be caused by #290

evanunderscore:
Feature and implementation look good to me. I'll leave it to @kislyuk to make the final call on whether to merge, and if so whether it warrants a spot in the README.

kislyuk:
LGTM. CI failure is unrelated, ignoring for now.

This will be effectively undocumented for now; @msc42, please consider contributing a documentation section on it.

evanunderscore:
This has a critical problem that I missed when reviewing it - there's currently only one copy of the completion function that gets registered, so whatever is registered most recently will win. This is enough to break it: https://github.com/evanunderscore/argcomplete/commit/8fb4dada0d378fe4ccf01b960b0c0a860e2b0a09

The simplest option I can see to make this work would be to dynamically generate the name of the completion function, as we're currently doing for fish.

PR #292, opened 2020-03-19
Remove expected test failure for new versions of fish

### LINKED ISSUE #290: fish 3.1: test_special_characters_double_quoted (test.test.TestFish) ... unexpected success
Since we have updated fish to 3.1 in Fedora, the package fails to build with:

```
test_special_characters_double_quoted (test.test.TestFish) ... unexpected success
```

I know how to fix this for us, by removing the test from:

https://github.com/kislyuk/argcomplete/blob/d2c12c3bfca2f1f11ad0737462360fba035fa71d/test/test.py#L1250-L1255

But I have no idea how to make it xfail only with fish 3.0.

evanunderscore:
Fixes #290


hroncok:
Thank you

PR #295, opened 2020-03-28
added dipslay completions test

volkov:
Test current behaviour for fixing help string for fish. See https://github.com/kislyuk/argcomplete/pull/264#discussion_r394792164 

Probably shouldn't be merged.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -506,6 +506,41 @@ class TestArgcomplete(unittest.TestCase):
         self.assertEqual("ttt", disp.get("--oh", ""))
         self.assertEqual("ccc", disp.get("--ch", ""))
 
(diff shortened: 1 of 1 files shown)


evanunderscore:
I think there's merit in merging this, then the difference in #264 will be more obvious.

test/test.py
```diff
@@ -506,6 +506,41 @@ def test_display_completions(self):
         self.assertEqual("ttt", disp.get("--oh", ""))
         self.assertEqual("ccc", disp.get("--ch", ""))
 
+        if not USING_PYTHON2:
```
evanunderscore: Since this doesn't share any code with the rest of the test, how about making it a separate test and adding e.g. `@skipIf(USING_PYTHON2, 'Subparser aliases aren't supported on Python 2')`?

PR #296, opened 2020-04-04
Improve auto completion for an arbitrary name using a given external script

msc42:
solves problems with pull request #288


evanunderscore:
Looks good, thanks!

Issue #166, opened 2016-12-07
Is it possible to support auto completion in PowerShell (PoSh)?


Paebbels:
Windows is shipped with PowerShell since years. The current version is 5.0 and the next release will be available on Linux and Mac OS too. It supports auto completion like Bash.

Is there any support for PowerShell planned?

(I could assist to implement it to some degree, but I can't offer a complete ready to use PR :).

kislyuk:
I'm open to it, but I don't use powershell, so you or someone else will have to come up with a PR.

Here is a good starting point: https://github.com/lzybkr/PSReadLine/issues/194

kislyuk:
Not interested in supporting PowerShell at this point.

aviita:
@kislyuk, so should it work in bash? Does not work for me in Ubuntu on WSL.

kislyuk:
@aviita yes, argcomplete works in bash. If you have problems making it work, please open a new issue and describe the full steps to reproduce.

aviita:
@kislyuk , 

Thanks. Sorry I failed to respond my self yesterday evening. I did get it working by installing azure cli again in bash. Also had to note that it only works with double tab press. Thanks.

Issue #267, opened 2019-09-11
git bash for Windows support


slanoe:
Is it possible to use argcomplete with git bash for Windows ?

kislyuk:
I'm not sure - I don't have it handy. What does `echo $BASH_VERSION` in that shell say?

slanoe:
```4.4.23(1)-release```

kislyuk:
That version has all the features necessary to support argcomplete. Are you having any specific issues using it?

slanoe:
It's OK with WSL
Thanks

evanunderscore:
I've revisited this and submitted draft PR #305 which provides some support for Git Bash.

Issue #142, opened 2016-09-22
Anyone got this working with bash from git for windows?


krist10an:
After some debugging, it seems it has issues opening fd 8 on my computer :
"Unable to open fd 8 for writing, quitting"

$ git --version
git version 2.10.0.windows.1

$ bash --version
GNU bash, version 4.3.46(2)-release (x86_64-pc-msys)

kislyuk:
Can you run the following on your Windows system and let us know the full output?
```
python -c 'import os; os.fdopen(8)'
python -c 'import os; os.fdopen(8, "w")'
python -c 'import os; os.fdopen(8, "wb")'
```

krist10an:
Sure:

```
$ python -c 'import os; os.fdopen(8)'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
OSError: [Errno 9] Bad file descriptor

$ python -c 'import os; os.fdopen(8,"w")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
OSError: [Errno 9] Bad file descriptor

$ python -c 'import os; os.fdopen(8,"wb")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
OSError: [Errno 9] Bad file descriptor

```

krist10an:
For reference I get the same result from Linux (Ubuntu 16.04,1 LTS x64, Python 2.7,.12):

```
vagrant@vagrant:~$ python -c 'import os; os.fdopen(8)'
python -c 'import os; os.fdopen(8, "wb")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
OSError: [Errno 9] Bad file descriptor
vagrant@vagrant:~$ python -c 'import os; os.fdopen(8, "w")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
OSError: [Errno 9] Bad file descriptor
vagrant@vagrant:~$ python -c 'import os; os.fdopen(8, "wb")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
OSError: [Errno 9] Bad file descriptor
```

evanunderscore:
You can only open that fd if you create it:

```
python -c 'import os; os.fdopen(8)' 8>&1
python -c 'import os; os.fdopen(8, "w")' 8>&1
python -c 'import os; os.fdopen(8, "wb")' 8>&1
```

Unfortunately this doesn't seem to work for me on Windows (in bash or cmd).

krist10an:
[...]

PR #302, opened 2020-05-19
Enable PowerShell in argcomplete

tibortakacs:
Currently, `argcomplete` does not support PowerShell (https://github.com/kislyuk/argcomplete/issues/166).

However, it is possible to write PowerShell script to enable `argcomplete`:
https://github.com/tibortakacs/powershell-argcomplete

This solution works well in PowerShell, however, it requires minor modification of the application which uses `argcomplete` (create special output stream, see at https://github.com/tibortakacs/powershell-argcomplete/blob/82c79a752dd0ebdeca5e7aa4e929f04253999363/mat.py#L92).

This is necessary because PowerShell does not support file descriptor '8'.

This PR would take a step forward, and enable `argcomplete` in PowerShell without any modification of the applications:
https://github.com/tibortakacs/powershell-argcomplete/tree/poc_add_ostream_logic_to_argcomplete

This approach introduces a new environment variable to specify the output stream's file descriptor. The PowerShell script could just set the value to '2' (stdout) which stream is used to gather completion results.

Looking forward to your opinion.

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -185,9 +185,10 @@ class CompletionFinder(object):
 
         if output_stream is None:
             try:
-                output_stream = os.fdopen(8, "wb")
+                output_stream_fd = int(os.environ.get("_ARGCOMPLETE_OSTREAM_FD", 8))
+                output_stream = os.fdopen(output_stream_fd, "wb")
             except:
-                debug("Unable to open fd 8 for writing, quitting")
+                debug("Unable to open fd {0} for writing, quitting".format(output_stream_fd))
                 exit_method(1)
 
         # print("", stream=debug_stream)
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -9,6 +9,7 @@ from .compat import USING_PYTHON2, str, sys_encoding, ensure_str, ensure_bytes
 from .completers import FilesCompleter, SuppressCompleter
 from .my_argparse import IntrospectiveArgumentParser, action_is_satisfied, action_is_open, action_is_greedy
 from .shell_integration import shellcode
+from ._mmap_stream_win import MappedMemoryStream
 
 _DEBUG = "_ARC_DEBUG" in os.environ
 
@@ -184,11 +185,26 @@ class CompletionFinder(object):
             debug_stream = sys.stderr
 
         if output_stream is None:
-            try:
+            output_stream_mode = os.environ.get("_ARGCOMPLETE_OSTREAM_MODE", "fd")
+
+            if output_stream_mode == "fd":
                 output_stream_fd = int(os.environ.get("_ARGCOMPLETE_OSTREAM_FD", 8))
-                output_stream = os.fdopen(output_stream_fd, "wb")
-            except:
-                debug("Unable to open fd {0} for writing, quitting".format(output_stream_fd))
+                try:
+                    output_stream = os.fdopen(output_stream_fd, "wb")
+                except:
+                    debug("Unable to open fd {} for writing, quitting.".format(output_stream_fd))
+                    exit_method(1)
+            
+            elif output_stream_mode == "mmf_win":               
+                output_stream_mmf_name = os.environ.get("_ARGCOMPLETE_OSTREAM_MMF_NAME")
+                if output_stream_mmf_name is None:
+                    debug("No _ARGCOMPLETE_OSTREAM_MMF_NAME, quitting.")
+                    exit_method(1)
+                
+                output_stream = MappedMemoryStream(output_stream_mmf_name)
+            
+            else:
+                debug("Invalid mode {}, quitting.".format(output_stream_mode))
                 exit_method(1)
 
         # print("", stream=debug_stream)
--- a/argcomplete/_mmap_stream_win.py
+++ b/argcomplete/_mmap_stream_win.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+
+import sys
+from ctypes import windll, cdll,\
+    c_char_p, c_wchar, c_size_t, c_ulonglong, c_wchar_p, c_void_p,\
+    sizeof,\
+    WinError
+from ctypes.wintypes import BOOL, DWORD, HANDLE, LPCWSTR, LPCVOID, LPVOID
+
+class MappedMemoryStream:
+    def __init__(self, mmf_name):
+        self._name = mmf_name
+
+    def write(self, data):
+        print("Python {:s} on {:s}".format(sys.version, sys.platform))
+
+        # Constants
+        FILE_MAP_ALL_ACCESS = 0x000F001F
+        SHMEMSIZE = 0x100
+
+        # Function definitions
+        kernel32_dll = windll.kernel32
+        msvcrt_dll = cdll.msvcrt
+
+        open_file_mapping_func = kernel32_dll.OpenFileMappingW
+        open_file_mapping_func.argtypes = (DWORD, BOOL, LPCWSTR)
+        open_file_mapping_func.restype = HANDLE
+
+        map_view_of_file_func = kernel32_dll.MapViewOfFile
+        map_view_of_file_func.argtypes = (HANDLE, DWORD, DWORD, DWORD, c_ulonglong)
+        map_view_of_file_func.restype = LPVOID
+
+        memcpy_func = msvcrt_dll.memcpy
+        memcpy_func.argtypes = (c_void_p, c_void_p, c_size_t)
+        memcpy_func.restype = LPVOID
+
+        unmap_view_of_file_func = kernel32_dll.UnmapViewOfFile
+        unmap_view_of_file_func.argtypes = (LPCVOID,)
+        unmap_view_of_file_func.restype = BOOL
+
+        close_handle_func = kernel32_dll.CloseHandle
+        close_handle_func.argtypes = (HANDLE,)
+        close_handle_func.restype = BOOL
+
+        get_last_error_func = kernel32_dll.GetLastError
+
+        # Connect to the memory mapped file
+        file_mapping_name_ptr = c_wchar_p(self._name)
+        mapping_handle = open_file_mapping_func(FILE_MAP_ALL_ACCESS, False, file_mapping_name_ptr)
+
+        print("Mapping object handle: 0x{:016X}".format(mapping_handle))
+        if not mapping_handle:
+            print("Could not open file mapping object: {:d}".format(get_last_error_func()))
+            raise WinError()
+
+        mapped_view_ptr = map_view_of_file_func(mapping_handle, FILE_MAP_ALL_ACCESS, 0, 0, SHMEMSIZE)
+        err = get_last_error_func()
+        print(err)
+
+        print("Mapped view addr: 0x{:016X}".format(mapped_view_ptr))
+        if not mapped_view_ptr:
+            print("Could not map view of file: {:d}".format(get_last_error_func()))
+            close_handle_func(mapping_handle)
+            raise WinError()
+
+        # Write data into the file
+        data_ptr = c_char_p(data)
+        byte_len = len(data)
+    
+        print("Message length: {:d} chars ({:d} bytes)".format(len(data), byte_len))
+
+        memcpy_func(mapped_view_ptr, data_ptr, byte_len)
+
+        # Detach
+        unmap_view_of_file_func(mapped_view_ptr)
+        close_handle_func(mapping_handle)
+
+    def flush(self):
+        # Just for the compatibility
+        pass
+
+
+# Just an example of usage    
+if __name__ == "__main__":
+    import os
+    import locale
+
+    msg = "Here is a new message."
+
+    # output_stream = os.fdopen(2, "wb")
+    output_stream = MappedMemoryStream("test_mmf")
+    output_stream.write(msg.encode(locale.getpreferredencoding()))
+    output_stream.flush()


evanunderscore:
The issue with stdout/stderr is they require the script's cooperation in not printing anything to them during argcomplete invocations, which is fine for scripts you write and maintain, but isn't part of argcomplete's general contract and can't be guaranteed to work with arbitrary argcomplete-compatible scripts.

Options for Windows support (in the context of Git Bash support) have been discussed in the past on #267 and #142, where @kislyuk suggested using WSL or investigating other IPC mechanisms on Windows. I was unable to find any IPC mechanisms that could be conveniently accessed from the command line, though I did not investigate Powershell at all.

If you are not able to find a Powershell-specific IPC mechanism, then the ability to opt-in to using stdout - with the associated caveats - seems like the next best option as far as I can see. Ultimately it is up to @kislyuk whether this feature is acceptable.

tibortakacs:
@evanunderscore Thank you for your response. You are right, this solution is not perfect because applications can use stdout before calling `autocomplete`. However, I believe this is discouraged by the pattern how `argcomplete` should be used.

Due to this, we can consider to change the approach: instead of passing file descriptor, we can make stdout usage explicit (_ARGCOMPLETE_OSTREAM_USE_STDOUT). It is less flexible but more obvious for the developers of the effect of the settings.

In the mean time, I have been working on some better IPC mechanism using memory mapped files, I am sure though it will be successful or not. I would surely require using Win32 DLLs (e.g. kernel32) with `ctypes` (https://docs.python.org/2/library/ctypes.html). Would be this an accepted option if it works?

@evanunderscore @kislyuk Looking forward to your opinion.

kislyuk:
Thanks for the thoughtful discussion.

I'm OK with parameterizing the fd in this way, as long as the default remains unchanged. I would *not* be OK using kernel32 via ctypes in argcomplete :)

With that said, in order for this PR to be merged, it would have to add appropriate documentation and specific instructions for how to use this functionality in PowerShell.

evanunderscore:
I've had another look into this, and managed to get a handle-based solution working. I've put up draft PR #305. Would something along these lines help you from PowerShell @tibortakacs?

tibortakacs:
@evanunderscore Thank you for the update. Not totally sure, I have to check whether the handle-based solution is supported by PowerShell or not. I am going to send feedback soon*.
(_*It might take a few days, I am OoO now. Sorry for the delay._)

evanunderscore:
While it would be ideal if PowerShell is able to perform the handle manipulation natively, the point of the Python wrapper on my branch is that it does the work for you. At the cost of running the interpreter one additional time, you get a program that outputs only the completions on stdout and nothing else. Interested in your opinion on whether this trade off is worthwhile.

tibortakacs:
@evanunderscore Thank you for the head-up. What do you mean "running the interpreter one additional time"? Could you please give me a short explanation how your solution is expected to work?

I have player with PowerShell and stdout handles. I am almost there, however, I found a small discrepancy yet. When I use GetStdHandle (https://docs.microsoft.com/en-us/windows/console/getstdhandle) in PowerShell to get the stdout handle, I always get `88`. When I pass it to the Python script, it works well, however, I cannot catch the stdout from the PowerShell script, it is always shown. The interesting point is that if I just run `h = msvcrt.get_osfhandle(1)` in Python, `h` handle value always changes, and its value is something like 3588, 3412 etc., but only for `1`. For `2` it is stable `92`. Going to contniue what is happening here, I just thought that you might have an idea here.

tibortakacs:
@evanunderscore The main issue can be demonstrated by using this example:
```Python
import os
import msvcrt

h = msvcrt.get_osfhandle(1)
print(h)
# Version 1 - Using h
# i = int(h)

# Version 2 - Using const value
i = int(84) 
if i == h:
    print("Equal")

stdout_fd = msvcrt.open_osfhandle(i, os.O_RDONLY | os.O_TEXT)

output_stream = os.fdopen(stdout_fd, "wb")
output_stream.write("Hello World!".encode())
output_stream.flush()
```

In case of Version 1:
```
(venv) PS> python .\failure.py
84
Hello World!
(venv) PS> python .\failure.py 1>$null
(venv) PS >$out = python .\failure.py
(venv) PS > $out
Hello World!2780```

So, the output can be properly redirected or captured. And it looks like that the value of `h` is `84`, but in case of `$out = python .\failure.py` `2780` is written into the output. Stange.

In case of Version 2:
```
(venv) > python .\failure.py
84
Equal
Hello World!
(venv) > python .\failure.py 1>$null
Hello World!
(venv) > $out = python .\failure.py
Hello World!
(venv) > $out
1916
```

So, the `output_stream` output is not part of the redirection or capute, but it is always shown by the console.

Do you have any idea what is going on here? :) If we can solve this mystery, there is not any blockage to move into the handle-based direction. Otherwise, I might merge the handle and the fd based solutions.

evanunderscore:
> What do you mean "running the interpreter one additional time"? Could you please give me a short explanation how your solution is expected to work?

You can see the edits I made to the shell functions on in my draft PR - rather than use the shell to run the command and redirect output (`"$@" 8>&1 9>&2 1>&9 2>&1`) the Python wrapper is called instead (`__python_argcomplete_run_subprocess "$@"`). The wrapper then handles these redirections internally.

Ordinarily, tab completion runs the shell function, which will run your application (via the Python interpreter) in order to get the completions. In this case, only one Python interpreter is launched.

With my proposed change, the shell function will run my Python wrapper (which requires you to run the interpreter) which then uses subprocess to run your application (via a completely separate Python interpreter). The "one additional time" I'm referring to is the step in the middle where we need to run my script. The performance impact is the time to start the interpreter and execute my script. Theoretically this need not be implemented in Python; we just need *something* to do the redirection.

> Do you have any idea what is going on here?

Handles are per-process opaque identifiers for resources. You shouldn't have any expectation that PowerShell's handle to stdout should match Python's handle to stdout, or even that of another instance of PowerShell.

In version 1, you're effectively just duplicating your own stdout FD and writing to it. Whether this means you're writing to PowerShell's stdout or somewhere else depends on whether PowerShell performed any redirection. I'm not certain why `2780` is printed, but it could be related to creating the file descriptor with `os.O_TEXT` but then opening it in binary mode.

In version 2, you're writing directly to PowerShell's stdout. It appears that the handles were inheritable, so you've been able to open them in the child process and then write directly. This is effectively the same thing that I'm doing in my solution: I'm [passing the handles for *the wrapper script's* stdout/stderr via environment variables](https://github.com/kislyuk/argcomplete/pull/305/files#diff-5ef453bdca37fe1586cc04e8604f3168R29-R33) that the child then [uses to open the output and debug streams](https://github.com/kislyuk/argcomplete/pull/305/files#diff-2ceaeff732e9e0c91ad323fd442cfee3R180-R192).

The difference is that the shell function is then capturing stdout from the wrapper Python process. The fact that this is happening in an intermediary script is what's making it work; if I were connected directly to Bash's stdout, capturing the output would not be possible.

In your case, you can use my script directly and I expect it should work for you. There may be a way in PowerShell to get a handle connected to some kind of pipe, then you could pass that handle and eliminate the need for the intermediate Python wrapper, though I don't know enough about PowerShell to suggest a starting point for that. If PowerShell scripts are run as separate processes, you might be able to get that to work, but it would only be worthwhile if PowerShell's startup cost is less than Python's.

evanunderscore:
> Handles are per-process opaque identifiers for resources. You shouldn't have any expectation that PowerShell's handle to stdout should match Python's handle to stdout, or even that of another instance of PowerShell.

I should probably walk this comment back a little bit, since I'm not absolutely certain of the details myself. Obviously if the handles are inherited then they will be the same. If you're seeing the stdout handle change but stderr remain the same, I'm guessing it's because you're redirecting stdout so the handle is referring to something else, while stderr remains connected to PowerShell's stderr.

tibortakacs:
@evanunderscore  Thank you for your very detailed response. I am going to process it, and send reponse soon.

@kislyuk Thank you for your feedback. I understand your concerns of the usage of 'kernel32' or any other Windows DLL. Please let me ask, just for my education, which are your main concerns regarding this question?

@kislyuk @evanunderscore Before the answer about usage of "kernel32" I had figured out how to use memory mapped files as IPC between PowerShell (and actually any win32 process) and Python, just needed some time to clean it and make it work with `argcomplete`. I have pushed a new commit with a proof of concept solution to show how it could work, it might be useful now or later.

Please find the corresponding PowerShell changes here:
https://github.com/tibortakacs/powershell-argcomplete/commit/c0a12f4fc65e0e37c27f27ef7e20b475a3853dbe

Since the `stdout` based solution worked properly well with PowerShell, I would consider the memory mapped based solution as a backup plan for applications which uses `stdout` before calling `argcomplete`. Therefore I have introduced the "output stream mode" concept: the default is the file descriptor based one to preserve the original behavior, and I have added the "mmf_win" one for the memory mapped file IPC. It could be added @evanunderscore 's handle based solution as well.

Looking forward to your opinion.

evanunderscore:
Great work getting the memory-mapped files working! In this case though I don't think it's the most appropriate tool for the job. Having to specify the size in advance and having it backed by the page file is not ideal. Ideally we want something entirely in-memory.

In case I wasn't clear before, I was suggesting that at a minimum you would be able to use the `__python_argcomplete_run_subprocess` script on my branch from PowerShell, in which case the fact that it passes handles becomes an implementation detail that you need not be aware of. Have you managed to get that working? I think it's a reasonable starting point.

However I think you can get PowerShell to do that work itself and eliminate the need for my script entirely. I've pushed https://github.com/evanunderscore/argcomplete/commit/6e04eb55402a67c3d1b1e8a9d27c250918c6f707 which modifies my wrapper to use a pipe instead of shortcutting directly to its own stdout, which you can use as a model for how I would do this in PowerShell. If you're able to call [CreatePipe](https://docs.microsoft.com/en-us/windows/win32/api/namedpipeapi/nf-namedpipeapi-createpipe) and make the write handle inheritable, you should be able to specify that handle as an environment variable and use my branch without modification.

If you can get that working, you shouldn't need any changes beyond what's on my branch. I am inexperienced with PowerShell, but if you run into trouble, let me know and I'll do my best to help out.

tibortakacs:
Thank you, @evanunderscore . I agree, the memory mapped solution is not best approach. It might be useful in future somewhere else :). Just for education: this page describes the concept: https://docs.microsoft.com/en-us/dotnet/standard/io/memory-mapped-files It does write that it uses virtual memory, however, the "Non-persisted memory-mapped files" which I used _"are suitable for creating shared memory for inter-process communications (IPC)."_ Do you have any more insight how it works behind the scene? My understanding was that it is purely memory-based solution as well. Just the put the last bricks to the right place in this topic :)

Thank you for your detailed description of the handles and file descriptors, and the solution you proposed. I have also done some research to understand the basics of them (this blog post is particulary useful: http://lackingrhoticity.blogspot.com/2015/05/passing-fds-handles-between-processes.html). And now I understand your PR which I believe is great, it really works out of box, I am happy to move into this direction in PowerShell.

Let me ask though how robust this solution is comparing the pure stdout based one? The pipe-based is just much better for sure because it solve the accidental (intentional?) stdout writing before argcomplete, however, it is also more complex with `msvcrt` call and handle inheritance. Just wondering whether there is any scenario which we can recall now where there is a potential risk that it would not work, and we could keep the file descriptor based as a backup?

Let me ask a small question regarding the final usage as well. You used `py.exe` which on my system is `C:\WINDOWS\py.exe`. I believe it would be better to call the interpreter which is used by calling `__python_argcomplete_run_subprocess`. Would it make sense to change it to `sys.executable`? (I know this is an implementation detail, however, for the project I have been working on would be an important point.)

evanunderscore:
I don't know for sure, but I'd suspect it uses [CreateFileMapping](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createfilemappinga) under the hood, which says this:

> If hFile is INVALID_HANDLE_VALUE, the calling process must also specify a size for the file mapping object in the dwMaximumSizeHigh and dwMaximumSizeLow parameters. In this scenario, CreateFileMapping creates a file mapping object of a specified size that is backed by the system paging file instead of by a file in the file system.

So my understanding is that it does create non-persisted shared memory, but that memory is actually backed by the system paging file.

> Just wondering whether there is any scenario which we can recall now where there is a potential risk that it would not work

It will work as long as the handles are inherited by the Python process argcomplete runs in. For example, [when the Windows launcher creates the process](https://github.com/python/cpython/blob/404b23b85b17c84e022779f31fc89cb0ed0d37e8/PC/launcher.c#L792-L793) it sets the [bInheritHandles](https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessw) parameter to TRUE. If this were to be changed to FALSE, the handle would not be inherited when our process is launched via py.exe. Similarly if some other step is introduced in the chain that does not allow handles to be inherited, this won't work. The same would be true on Linux if there were something in the chain that didn't allow the file descriptors to be inherited.

> I believe it would be better to call the interpreter which is used by calling __python_argcomplete_run_subprocess

The fact that this script is written in Python is completely an implementation detail. It should not matter at all which interpreter is used to run it (though I haven't specifically tested against Python 2). There is absolutely no reason to assume that the interpreter used to run that script should be the same as the child (which could have a shebang line indicating it is to be run with a completely different interpreter). This is exactly the reason I'm using py.exe; because I want the target script to run with the interpreter *it* specifies, not the interpreter my script specifies.

The reason I have to do this is described in a comment above where I'm running py.exe, but in short: I don't actually know how the shell is going to run the script. I also can't guarantee that using py.exe will produce the same result, but it's the best method I could come up with. There may be a better alternative that involves launching another Bash shell to execute the command.

However none of this should matter to you if you implement what the script is doing directly in PowerShell. In that case, you *are* the shell; you can just run the command unmodified.

tibortakacs:
@evanunderscore Thank you for your detailed responses again.

Regarding the shared memory question, interesting point, .NET does use this method under the hood. I can imagine that the .NET documentation and the CreateMappedFile documentation can be aligned, but this is definitely out of scope of this PR.

Thank you for your analysis regarding the robustness of the pipe-based solution. I am convinced this is the right way to go to.

About my interpreter question: I see now how "nooby" my question was :), thank you for your detailed response. I learned now about `py.exe` in Windows, and this is the right solution.

>However none of this should matter to you if you implement what the script is doing directly in PowerShell. In that case, you are the shell; you can just run the command unmodified.

I think the PowerShell script would just call `py.exe __python_argcomplete_run_subprocess` as you mentioned above. In the current state, I do not see too big penalty of running two Python interpreters, and using the `__python_argcomplete_run_subprocess` would align PowerShell to other shells.

Would you agree with this direction?

tibortakacs:
@evanunderscore @kislyuk Thanks to @evanunderscore 's change, the PowerShell support does not need any code change, it only needs the wrapper script. Therefore, the scope of this PR is mainly adocumentation change. Which way do you prefer?

1. Keeping the https://github.com/tibortakacs/powershell-argcomplete repository with the detailed documentation the PowerShell wrapper script, and adding reference with short summary here.

2. Moving the detailed documentation here.
a. Create a new `POWERSHELL.rst` page with a reference in `README.rst`.
b. Just extend `README.rst`.

3. Any combinations of the aboves.

Please let me know which way you prefer.

evanunderscore:
> About my interpreter question: I see now how "nooby" my question was :), thank you for your detailed response. I learned now about py.exe in Windows, and this is the right solution.

No problem. It was a perfectly valid question, and one that has come up several times before in this project, since using `sys.executable` is *usually* the right way to do things. However in this case it is not, and I wanted to be very clear about why.

Your implementation looks good! I don't expect you should have any need to use the wrapper script at all; I would recommend only including the handle-based version unless someone can come up with a compelling reason otherwise.

One thing to mention - you shouldn't have to use `py.exe` at all in your code; when the user runs `./mat.py`, Windows will use the file association for `.py` files to work out how to launch the script. In most cases this will be `py.exe`, but not always.

kislyuk:
We do not officially support Powershell, but argcomplete contributors have since been able to achieve unofficial support using other approaches. I'm going to go ahead and close this PR - it has no chance of being merged in its current state, and we will not be using memory mapping or ctypes in argcomplete.

PR #305, opened 2020-05-23
Add support for script completion on Windows via Git Bash

evanunderscore:
* How does this impact WSL users?
    * Is the OS environment variable set there?
    * Should we make this feature opt-in instead?
* Global completion of python/py does not work
    * Should be fixable with a flag to tell the script not to launch via py.exe
* Completion of pip-generated script wrappers does not work
    * Much harder; existing logic parses the wrapper to find the correct interpreter to then find the script to check for the argcomplete marker
    * Alternatively we could just guess an interpreter and hope for the best

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -177,18 +177,41 @@ class CompletionFinder(object):
             # not an argument completion invocation
             return
 
+        stdout_fd = 8
+        stderr_fd = 9
+
+        def open_osfhandle(handle):
+            import msvcrt
+            return msvcrt.open_osfhandle(handle, os.O_RDONLY | os.O_TEXT)
+
+        stdout_handle = os.environ.get("_ARGCOMPLETE_STDOUT_HANDLE")
+        if stdout_handle is not None:
+            stdout_fd = open_osfhandle(int(stdout_handle))
+        stderr_handle = os.environ.get("_ARGCOMPLETE_STDERR_HANDLE")
+        if stderr_handle is not None:
+            stderr_fd = open_osfhandle(int(stderr_handle))
+
         global debug_stream
         try:
-            debug_stream = os.fdopen(9, "w")
+            debug_stream = os.fdopen(stderr_fd, "w")
         except:
             debug_stream = sys.stderr
 
         if output_stream is None:
             try:
-                output_stream = os.fdopen(8, "wb")
+                output_stream = os.fdopen(stdout_fd, "wb")
             except:
-                debug("Unable to open fd 8 for writing, quitting")
+                debug("Unable to open fd {} for writing, quitting".format(stdout_fd))
                 exit_method(1)
+
+        # Almost always in the middle of a user typing a line;
+        # move to the next line before any other debug output.
+        debug()
+
+        if stdout_handle is not None:
+            debug("Mapped stdout handle {} to fd {}".format(stdout_handle, stdout_fd))
+        if stderr_handle is not None:
+            debug("Mapped stderr handle {} to fd {}".format(stderr_handle, stderr_fd))
 
         # print("", stream=debug_stream)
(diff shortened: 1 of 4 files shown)


evanunderscore:
Examples using Git Bash on Windows. I pip-installed argcomplete into a virtual environment and ran this from the source tree.

```
$ eval "$(scripts/activate-global-python-argcomplete --dest=-)"
$ test/prog
basic   break   cont    debug   env     -h      --help  point   quote   space   spec
```

```
$ eval "$(scripts/register-python-argcomplete test/prog)"
$ test/prog
basic   break   cont    debug   env     -h      --help  point   quote   space   spec
```

tibortakacs:
@evanunderscore As we talked in the PowerShell PR (https://github.com/kislyuk/argcomplete/pull/302), your solution is great to support PowerShell auto-completion. Can I ask what is the schedule for this PR? Just asking which documentation I should prepare for the PowerShell support. Thank you in advance.

evanunderscore:
@tibortakacs I've ended up going with a different solution for Git Bash in #307, however we could certainly still add support for handles for use with PoweShell. Once #307 is merged, I can submit a new PR to do that.

tibortakacs:
@evanunderscore Thank you. I see that you have implemented a temporary file based solution. I am just wondering that it would be also an option to PowerShell as well. Can I ask for my education what the background is, why is the temporary file solution better than the previous pipe based (in-memory) solution? Which way would you prefer for PowerShell?

evanunderscore:
@tibortakacs See the discussion on #306 for why a temporary file is more appropriate for Git Bash. In short, there is no convenient way to manipulate an inheritable pipe from Git Bash without launching another process of some description. The wrapper script had several drawbacks that outweighed the untidiness of using temporary files.

My preference is still to use the handle-based solution for PowerShell, but the downside is that introduces a code path that cannot be tested by the CI; we'd be relying on user reports to tell us whether we accidentally break functionality.

tibortakacs:
@evanunderscore I hope you are well. Can I ask what is your preferred schedule of merging this change? Just asking in because I am getting closer to the point where it would be great if I could use `argcomplete` in PowerShell, and your handle-based solution is best for this purpose. Please let me know if I can do anything which supports you here. Thank you in advance.

evanunderscore:
@tibortakacs I have no set schedule; I was planning to submit this change separately once @kislyuk has had a change to review and merge #307. The handle change is no longer necessary for Git Bash so will be added purely for PowerShell.

tibortakacs:
@evanunderscore Thank you, looking forward to it! In the mean time, I will finish the documentation PR for the standard output based and the handle based solutions.

evanunderscore:
@tibortakacs I've created a separate branch with just the handle based changes at https://github.com/evanunderscore/argcomplete/tree/powershell-handles. It makes sense to me to include that change in the same branch that adds the documentation for how to use it with PowerShell, since that's currently the only use case. Are you able to incorporate the changes into your branch?

tibortakacs:
@evanunderscore Absolutely, thank you. Going to do that soon (most probably next week), and let you know the result.

tibortakacs:
@evanunderscore Other question about `git`. I was thinking that we might be able to enable autocomplete `git` in PowerShell as well. At least it would be an interesting excercise for me :). I could not find, however, any context how `git` uses `argcomplete`. Could you give me some starting hints which can drive me forward? Thanks a lot!

evanunderscore:
As far as I'm aware, Git doesn't use argcomplete at all; it looks like it's been done by hand: https://github.com/git/git/blob/master/contrib/completion/git-completion.bash

tibortakacs:
@evanunderscore Ah, thank you. I misunderstood the title of PR. This is about to support scripts in "Git Bash on Windows" which are based on `argcomplete` but _not_ `git` itself. Thank you for the clarification.

evanunderscore:
I'm closing this since Git Bash is supported via temporary files as described above, and I believe the handle-based solution mentioned in [this comment](https://github.com/kislyuk/argcomplete/pull/305#issuecomment-655233998) will be included in PowerShell changes.

PR #306, opened 2020-05-26
Add partial support for script completion on Windows via Git Bash

evanunderscore:
Supersedes #305 

This should have no observable impact unless the user explicitly opts in to this change by setting ARGCOMPLETE_WINDOWS_COMPAT, meaning that existing users using Bash with WSL will not see any change.

I believe this is compatible with the approach I described in #302, where PowerShell could pass a pipe handle via _ARGCOMPLETE_STDOUT_HANDLE. There is actually no need to do the same for the debug stream, since argcomplete already falls back on stderr.

@kislyuk Are you happy with this approach? Happy to answer questions if any of my code or comments are unclear.

### DIFF
--- a/scripts/__python_argcomplete_run_windows
+++ b/scripts/__python_argcomplete_run_windows
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+"""
+Script mimicking __python_argcomplete_run using subprocess.
+
+Runs a Python script as a subprocess, printing completions on stdout
+and any other output on stderr.
+
+Called by the shell wrappers when ARGCOMPLETE_WINDOWS_COMPAT
+is defined in the environment.
+
+Useful in Windows environments where the file descriptor manipulation
+required by argcomplete is not possible from shell scripts,
+for example with a typical install of Git Bash.
+
+This script uses msvcrt; it is not suitable for use on non-Windows platforms.
+"""
+import os
+import msvcrt
+import subprocess
+import sys
+
+assert 'ARGCOMPLETE_WINDOWS_COMPAT' in os.environ
+
+# Inform the child of stdout/stderr handles via the environment.
+# These handles are inherited by setting close_fds=False.
+# They appear to be inheritable by default;
+# alternatively we could use os.set_handle_inheritable().
+# We use handles instead of file descriptors
+# because they are more easily inherited on Windows.
+# Even if we create the file descriptors, make them inheritable
+# and use os.spawn* instead of subprocess so that the
+# file descriptors are inherited by py.exe,
+# py.exe won't pass them to the Python process it spawns anyway.
+# See also: https://www.python.org/dev/peps/pep-0446/#inheritance-of-file-descriptors-on-windows
+stdout_handle = msvcrt.get_osfhandle(1)
+stderr_handle = msvcrt.get_osfhandle(2)
+env = os.environ.copy()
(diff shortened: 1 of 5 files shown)


scripts/__python_argcomplete_run_windows
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
```
kislyuk, line 1: I'm not comfortable with installing this script standalone on non-Windows platforms. Can we use setuptools conditionals to only install it on Windows?

evanunderscore: As far as I'm aware, there's nothing built in to setuptools that would let you conditionally include/exclude scripts (though I'm happy to be corrected). You could conditionally modify the thing you're passing to scripts in setup.py, but that would only work when installing from an sdist; for a wheel, which pip prefers, the decision is made at build time, meaning the script would be excluded globally if built on Linux and included globally if built on Windows. If we were to do that, we would have to update the instructions for Windows users telling them to `pip install argcomplete --no-binary argcomplete` so that they actually get the helper script.

I didn't take the approach of including the script under the argcomplete package and invoking it with `python -m argcomplete.__python_argcomplete_run_windows` for a couple of reasons. Firstly I don't necessarily know which interpreter I should be using to run the script, and secondly argcomplete does the majority of its work in its `__init__.py`, meaning running it that way would needlessly import all of argcomplete's dependencies, which negatively impacts response time.

So it seems our options are:
* Include the script unconditionally (it is useless on non-Windows platforms but also harmless; it will exit with an ImportError when it attempts to import msvcrt)
* Conditionally include the script in setup.py, advise Windows users to install from sdist
* Include only the handle writing changes, maintain the wrapper script in a separately installable package

I know this has been discussed several times before, but I'd like to take a moment to look back at the [tempfile IPC approach](https://github.com/evanunderscore/argcomplete/commit/481308a48efde8c8afd9aca4fef73033400e8d4a), which does have quite a few advantages:
* Does not require a helper script, so this packaging issue goes away
* Runs faster, since you don't need to launch an extra Python interpreter
* Does not need to guess using py.exe, since the shell launches the child directly
* No Windows-specific code is required, meaning this could be tested by Travis

Ultimately the decision is yours, so let me know how you would like to proceed with this.

(Irrespective of your decision here, I still believe directly passing a handle is the best option for PowerShell, since it has the capacity to create a pipe and provide the handle, whereas Git Bash does not.)

kislyuk: Thanks for analyzing the trade-offs clearly.

I agree that the tempfile IPC approach is preferable in light of this. We should be able to activate tempfile IPC only on Windows, so that gets rid of the issue here.

Does this sound reasonable?

kislyuk: (The main concern to watch out for with tempfiles is privacy issues from tempfiles being left behind; but I think given everything we discussed and given the scope limitation to Windows only, we should be able to mitigate that)

evanunderscore: I was intending to make temp files explicitly opt-in via an environment variable, but I can have it work by default on Windows if you'd prefer a more out of the box solution.

Creating the temporary files using `mktemp` should mean the files are only readable by the user (and a superuser) on both Windows and Linux, which means this won't be any less secure than your history file.

I'll prepare the other branch as a PR and we can continue the discussion there.

PR #311, opened 2020-07-16
Allow overwriting CompletionFinder debug_stream

crobinso:
We use argcomplete for `virt-install` (part of [virt-manager](https://github.com/virt-manager/virt-manager)) and it works great for us!

virt-install command line syntax is quite large and we have
many argument substrings, so we implement custom completer
functions. Naturally we want to test this as well, so we have
some unittest infrastructure for hitting this code. We capture
output by overriding the CompletionFinder output_stream=
argument.

I'm working on moving virt-manager to use pytest as our test
runner, but our completion tests were triggering weird pytest
behavior. After the first completer test was run the pytest
debug logs would show lots of 'Bad file descriptor' errors
and no more test results would be reported.

This seems to have something to do with CompletionFinder
debug_stream behavior, which will unconditionally attempt to
interact with fd=8 if it is available. For the shell case
this is desired behavior but maybe not for the unittest case.

This small series adds a debug_stream option similar to
output_stream, to allow API users to override the debug_stream
default behavior. I verified this can be used to avoid the
issues I'm seeing with pytest.

Patch #1 renames the global debug_stream, so we can use that
variable name for the __call__ signature.

Patch #2+#3 fix some doc issues in this area.

Patch #4 adds debug_stream argument and test coverage

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -147,6 +147,8 @@ class CompletionFinder(object):
             Method used to stop the program after printing completions. Defaults to :meth:`os._exit`. If you want to
             perform a normal exit that calls exit handlers, use :meth:`sys.exit`.
         :type exit_method: callable
+        :param output_stream: Stream object to write output to, or None for default behavior.
+        :type output_stream: file-like object or None
         :param exclude: List of strings representing options to be omitted from autocompletion
         :type exclude: iterable
         :param validator:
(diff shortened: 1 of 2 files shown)


crobinso:
FWIW the travis failure seems unrelated to this PR.

kislyuk:
Thanks for your contribution. I'm afraid this level of configurability is outside the scope of argcomplete's commitment to its users and we will not be able to take on this feature and its associated future maintenance overhead.

If you need something like this for testing, I recommend looking into subclassing, mocking, or monkey-patching to inject this behavior in your test suite.

PR #313, opened 2020-07-23
Bash nounset mode fixes

scop:


evanunderscore:
Can you add a test case for this please? Adding it to this base class will test both the command completion and global completion:  https://github.com/kislyuk/argcomplete/blob/86fb4be716e6d6fc914df83f8553e6675862a2dd/test/test.py#L1185

Something like this should work:

```
def test_nounset(self):
    self.sh.run_command('set -o nounset')
    self.test_simple_completion()
```

scop:
Sure, done.

evanunderscore:
CI failure seems unrelated, I'll have to look into that when I get time. Thanks!

PR #314, opened 2020-09-07
Xonsh support

### LINKED ISSUE #271: 🐚 xonsh support
Hi! Thank you for argcomplete! It is great!

I use it in bash but now trying xonsh. Could you please help me to add [xonsh](https://xon.sh/) support? Xonsh has [Programmable Tab-Completion](https://xon.sh/tutorial_completers.html) for it and it's very short to understand.

1. Here the sample of function that calls when user type `<TAB>` anywhere and see `item1 item2` always:
```python
def sample_completer(prefix, line, begidx, endidx, ctx):
    '''
    prefix: the string to be matched (the last whitespace-separated token in the current line)
    line: a string representing the entire current line
    begidx: the index at which prefix starts in line
    endidx: the length of the prefix in line
    ctx: the current Python environment, as a dictionary mapping names to values
    return: options
    '''
    return {"item1", "item2"}
```

2. If user type 
```
python sample.py --argu<TAB>
                 ^17   ^23
``` 
then:
```python
prefix = "--argu"
line = "python sample.py --argu"
begidx = 17
endidx = 23
```

3. If you want to play with it just get the [xonsh.AppImage](https://github.com/niess/linuxdeploy-plugin-python/releases/download/continuous/xonsh-x86_64.AppImage) (`chmod +x xonsh.AppImage`) or `pip install xonsh` and run (just copy and paste in xonsh):
```
def sample_completer(prefix, line, begidx, endidx, ctx):
    return {"item1", "item2"}
	
completer remove sample_completer && completer add sample_completer sample_completer 
```

Could you please advise the best way to add argcomplete to xonsh?

If you can imagine some hacks like call `bash -c 'argcomplete(line)'` inside `sample_completer` it also would be interesting :)

anki-code:
Hello! Thank you for argcomplete! It's useful!

There is extension for argcomplete in xonsh. This PR adds info about this extension.

Screenshot_20200908_123219

Thanks!

Closes #271 and https://github.com/xonsh/xonsh/issues/1409

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,7 +6,7 @@ Argcomplete provides easy, extensible command line tab completion of arguments f
 
 It makes two assumptions:
 
-* You're using bash as your shell (limited support for zsh, fish, and tcsh is available)
+* You're using bash as your shell (limited support for zsh, fish, tcsh and xonsh is available)
 * You're using `argparse <http://docs.python.org/3/library/argparse.html>`_ to manage your command line arguments/options
 
 Argcomplete is particularly useful if your program has lots of options or subparsers, and if your program can
--- a/README.rst
+++ b/README.rst
@@ -292,6 +292,13 @@ You can disable this feature by removing ``_ARGCOMPLETE_DFS`` variable, e.g::
 
     register-python-argcomplete --shell fish my-awesome-script | grep -v _ARGCOMPLETE_DFS | .
 
+Xonsh Support
+-------------
+To activate completions for xonsh use `xontrib-argcomplete <https://github.com/anki-code/xontrib-argcomplete>`_::
+
+    xpip install xontrib-argcomplete
+    xontrib load argcomplete
+
 Git Bash Support
 ----------------
 Due to limitations of file descriptor inheritance on Windows,


anki-code:
@kislyuk please take a look :)

evanunderscore:
My two cents - I think it's a good idea to include the known argcomplete extensions somewhere, but there should be some clear delineation between what is supported and maintained within argcomplete and what is provided by third party packages. Ultimately it's up to @kislyuk though.

kislyuk:
Thank you for your PR. It cannot be accepted in its current form because the argcomplete team does not support this external  package, and its installation instructions are outside the scope of this project's documentation.

If you want, you can open a new PR to add a link under the **Links** section in the argcomplete Readme instead.

anki-code:
@kislyuk what if I create pull request where xontrib (xonsh extension) will be in argcomplete repo? In this case xontrib will be installed with argcomplete (`pip install argcomplete` command also will add xonsh argcomplete extension to `xontrib` directory) - this way argcomplete team can maintain it.

kislyuk:
Sorry, I'm not interested in maintaining this extension in this repo.

anki-code:
@kislyuk thank you for merge PR with the link. I'm just interesting. I see that the Fish support is realised by another user and you merged it. You're not the Fish support developer as I see. 

Here we have the similar situation. I have clear concept of extension and xonsh has clear concept of support this type of extensions. The installation process via PyPi makes the solution seamless. No crutches in this process.

I could be support the extension as Fish and Zsh supporters doing here.

But you reject the proposal. I'm wondering why.

kislyuk:
I consider fish and zsh to be worth supporting, and I do not consider xonsh worth supporting at this time.

PR #335, opened 2020-11-29
Print `warn()` message from beginning of line

### LINKED ISSUE #334: `warn()` message incorrectly indented one space character
The current implementation of `warn()` annoyingly indents its message by one space character.

For example, the following:

```py
warn("Log in to X to auto-complete Ys")
```

renders

```pycon
 Log in to X to auto-complete Ys
^ <- space character
```

but it should be

```pycon
Log in to X to auto-complete Ys
^ <- no leading space
```

**Why is this happening?**
This is due to the fact that `print()` (which is what `warn()` uses under the hood) uses space (' ') as its separator. Hence, since a newline ('\n') is printed first, a space will be inserted before the next item (the actual message) is output.

**Proposed solution**
Simply do two `print()` calls. One for prepending the newline, and another for the message. The first `print()` doesn't even need a newline since `print()` automatically outputs a '\n' at the end.

In code:

```py
def warn(*args):
	print()
	print(file=debug_stream, *args)
```

I'll throw together a PR, if you come to agree with this being the expected behavior.

ulken:
The previous behavior was to indent the message by one space character.

Fixes #334

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -683,6 +683,5 @@ def warn(*args):
     """
     # Don't be tempted to use `print("\n",..., *args)`,
     # as that will indent **args** by one space character
-
     print(file=debug_stream)
     print(file=debug_stream, *args)
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -681,4 +681,7 @@ def warn(*args):
     Prints **args** to standard error when running completions. This will interrupt the user's command line interaction;
     use it to indicate an error condition that is preventing your completer from working.
     """
-    print("\n", file=debug_stream, *args)
+    # Don't be tempted to use `print("\n",..., *args)`,
+    # as that will indent **args** by one space character
+    print()
+    print(file=debug_stream, *args)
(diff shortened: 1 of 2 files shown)


ulken:
I'm working on adding a test, to ensure the issue doesn't accidentally pop up again. Having a little trouble capturing output from the `debug_stream`. I'll give it a little more effort, then I'll have to throw in the towel on this one.

ulken:
> I'm working on adding a test, to ensure the issue doesn't accidentally pop up again. Having a little trouble capturing output from the `debug_stream`. I'll give it a little more effort, then I'll have to throw in the towel on this one.

Got it working. Might not be the prettiest solution, but does the job. Feel free to adjust/remove the test if you disagree.

ulken:
@kislyuk All yours.

kislyuk:
Looks great, thanks.

ulken:
@kislyuk Noticed `2.7` is failing. Sorry, I didn't think to test with it (don't have it installed atm). I'm pretty sure it's due to incompatibility of `StringIO`. I'll try to fix it and submit a PR.

ulken:
Once fixed, will you release a new patch version with the fix any time soon, you think?

PR #338, opened 2021-01-12
Fix fish support

kenji-miyake:
Although fish-shell support was implemented in https://github.com/kislyuk/argcomplete/pull/260, I think some parts are wrong.

1. Probably [if-condition for _ARC_DEBUG in shell_integration.py](https://github.com/kislyuk/argcomplete/blob/d4197606e3b342d49ec4a9ed51f2fe89a603b3f0/argcomplete/shell_integration.py#L70-L74) is reversed
    - It seems to output to `/dev/null` when `_ARC_DEBUG` **is set**.
2. [Fish Support section in README.rst](https://github.com/kislyuk/argcomplete#fish-support) is different from [comments in register-python-argcomplete](https://github.com/kislyuk/argcomplete/blob/d4197606e3b342d49ec4a9ed51f2fe89a603b3f0/scripts/register-python-argcomplete#L23)
    - README
        - `register-python-argcomplete --shell fish ~/.config/fish/completions/my-awesome-script.fish`
    - register-python-argcomplete
        - `register-python-argcomplete --shell fish my-favourite-script.py > ~/.config/fish/my-favourite-script.py.fish`

@kislyuk @volkov Could you please take a look? Thank you.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -280,7 +280,7 @@ To activate completions for fish use::
 
 or create new completion file, e.g::
 
-    register-python-argcomplete --shell fish ~/.config/fish/completions/my-awesome-script.fish
+    register-python-argcomplete --shell fish my-awesome-script > ~/.config/fish/completions/my-awesome-script.fish
 
 Completion Description For Fish
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/argcomplete/shell_integration.py
+++ b/argcomplete/shell_integration.py
@@ -68,9 +68,9 @@ function __fish_%(executable)s_complete
     set -x COMP_POINT (string length (commandline -cp))
     set -x COMP_TYPE
     if set -q _ARC_DEBUG
-        %(argcomplete_script)s 8>&1 9>&2 1>/dev/null 2>&1
-    else
         %(argcomplete_script)s 8>&1 9>&2 1>&9 2>&1
+    else
+        %(argcomplete_script)s 8>&1 9>&2 1>/dev/null 2>&1
     end
 end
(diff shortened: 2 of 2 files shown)


volkov:
Looks good! There was bugs.

volkov:
I think there is some difference in bash and fish around this place

When I run `echo test 8>&1 9>&2 1>/dev/null 2>&1` in fish I see `test` in console output.
When I run same command in bash there is no output.

evanunderscore:
The changes look correct to me. `set -q` returns the number of queried variables that were undefined, i.e. a return code of 0 means the variable was defined. I'm not able to reproduce the issue @volkov is seeing; `echo test 8>&1 9>&2 1>/dev/null 2>&1` correctly outputs nothing for me (fish version 2.7.1).

evanunderscore:
Thanks for the PR!

volkov:
> I'm not able to reproduce the issue @volkov is seeing; echo test 8>&1 9>&2 1>/dev/null 2>&1 correctly outputs nothing for me (fish version 2.7.1).

Look's like it's only 3.1.2 feature (or bug), also can't reproduce on 2.7.1, report it as https://github.com/fish-shell/fish-shell/issues/7646

evanunderscore:
Thanks for reporting that issue @volkov, if it turns out more changes are required in argcomplete, feel free to open a new issue and/or PR to address it.

evanunderscore:
It looks like this issue may not be fixed any time soon in fish, so in the meantime you could choose to use the tempfile-based IPC mechanism we included for Git Bash on Windows: https://github.com/kislyuk/argcomplete/blob/9e2c6712c761d05f64f4d558a381d9be9fa13988/README.rst#git-bash-support

PR #340, opened 2021-01-22
Test that completer tracebacks are suppressed

evanunderscore:


evanunderscore:
This doesn't actually appear to capture the exception output.

PR #357, opened 2021-10-24
Replace assertRaisesRegexp with assertRaisesRegex

hugovk:
Deprecated since Python 3.2: 

* https://docs.python.org/3.10/library/unittest.html#deprecated-aliases

Removed in Python 3.11:

* https://docs.python.org/3.11/whatsnew/3.11.html#removed

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -1047,7 +1047,7 @@ class TestCheckModule(unittest.TestCase):
 
     def test_not_package(self):
         self._mkfile('module.py')
-        with self.assertRaisesRegexp(Exception, 'module is not a package'):
+        with self.assertRaisesRegex(Exception, 'module is not a package'):
             _check_module.find('module.bad')
         self.assertNotIn('module', sys.modules)
 


volkov:
Looks like 2.7 tests are broken after this fix

See https://github.com/kislyuk/argcomplete/runs/4130593269?check_suite_focus=true
```
  File "./test/test.py", line 1050, in test_not_package
    with self.assertRaisesRegex(Exception, 'module is not a package'):
AttributeError: 'TestCheckModule' object has no attribute 'assertRaisesRegex'
```

hugovk:
So it is, I missed that as the CI didn't for this PR.

Should be easy enough to fix, I'll make a new PR.

But I should ask, Python 2.7 is nearly two years past EOL, is now a good time to drop it?

Here's the pip installs for argcomplete from PyPI for October 2021, showing a very low proportion for 2.7: 

| category | percent |  downloads |
|:---------|--------:|-----------:|
| 3.7      |  83.67% | 18,027,112 |
| 3.8      |   5.84% |  1,258,254 |
| 3.6      |   4.27% |    919,956 |
| 3.9      |   2.61% |    561,976 |
| 3.5      |   1.62% |    349,570 |
| 2.7      |   0.97% |    208,482 |
| null     |   0.68% |    145,460 |
| 3.10     |   0.34% |     74,291 |
| 3.4      |   0.00% |        603 |
| 3.11     |   0.00% |         79 |
| 3.3      |   0.00% |         10 |
| 3.2      |   0.00% |          6 |
| Total    |         | 21,545,799 |

Source: `pip install -U pypistats && pypistats python_minor argcomplete --last-month`

hugovk:
Found https://github.com/kislyuk/argcomplete/issues/282#issuecomment-633142183:

> We are not actively maintaining support for Python 2, and I expect to drop all Python 2-specific compatibility code sometime later this year. In general we will follow EOL announcements from upstream to decide when to drop support, and in case of Python 2.7, EOL was January 1, 2020.

So looks like a 👍 to drop EOL versions. I'll put together a PR for that instead!

hugovk:
PR https://github.com/kislyuk/argcomplete/pull/360 restores support Python 2.7.

---

I had a go at dropping 2.7, but there's something strange going on with the CI with macOS: the 3.x jobs somehow have 2.7 in use but I couldn't quite figure out what.

For example, see https://github.com/hugovk/argcomplete/commit/e7c7ee707f9a0854b9fa1ce40bd4df9a26efc59b which just removes `from __future__ import print_function`, but the 3.x tests fail with a Python 2.7 `SyntaxError`:

```
-     print(file=debug_stream, *args)
-               ^
- SyntaxError: invalid syntax
```

https://github.com/hugovk/argcomplete/runs/4131846081?check_suite_focus=true

The Ubuntu tests are fine.

evanunderscore:
OSX bug aside, we still indicate that we support Python 2 [here](https://github.com/kislyuk/argcomplete/blob/e059d4f7bab1d6c3c903bdbefce0d7230740ec55/README.rst#python-support) and [here](https://github.com/kislyuk/argcomplete/blob/e059d4f7bab1d6c3c903bdbefce0d7230740ec55/setup.py#L47-L49). In my opinion, dropping support for Python 2 would require a major version bump for argcomplete, meaning Python 2 users can pin to `argcomplete<2`. Are you okay with this @kislyuk?

kislyuk:
@evanunderscore no, I would have to disagree on a major version bump being necessary to drop Python 2.7 support. You are correct that we need to remove the Python 2.7 classifier from the package classifiers. Using the setuptools `python_requires` option (https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) might also be necessary.

kislyuk:
@hugovk thanks so much for your effort to drop 2.7! I will remove all `from __future__` imports and troubleshoot any issues that we end up seeing.

kislyuk:
@evanunderscore actually looking more at the state of things, I think I agree we are due for a major version bump to argcomplete v2.0, with #352 merged and other changes in #361, and a migration to flit/pyproject.toml based build process that I've been planning.

hugovk:
> Using the setuptools `python_requires` option (https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires) might also be necessary.

Yes, it's a good idea to always have this (and it's been standardised by a PEP so other build backends beyond setuptools use it too, like flit).

Issue #327, opened 2020-10-29
fish-shell: function name is broken for local files (./script.py)


MaZderMind:
When Using argcomplete under Fish-Shell, the generated Code contains the literal Path to the Script as part of a function-name:

```
~/C/makefile-style-python register-python-argcomplete --shell fish ./foo/bar/baz

function __fish_./foo/bar/baz_complete
    set -x _ARGCOMPLETE 1
    set -x _ARGCOMPLETE_DFS \t
    set -x _ARGCOMPLETE_IFS \n
    set -x _ARGCOMPLETE_SUPPRESS_SPACE 1
    set -x _ARGCOMPLETE_SHELL fish
    set -x COMP_LINE (commandline -p)
    set -x COMP_POINT (string length (commandline -cp))
    set -x COMP_TYPE
    if set -q _ARC_DEBUG
        ./foo/bar/baz 8>&1 9>&2 1>/dev/null 2>&1
    else
        ./foo/bar/baz 8>&1 9>&2 1>&9 2>&1
    end
end
complete -c ./foo/bar/baz -f -a '(__fish_./foo/bar/baz_complete)'
```

`__fish_./foo/bar/baz_complete` is not a valid function name in Fish and thus `| source`ing the generated Snippet results in an Error:

```
~/C/makefile-style-python register-python-argcomplete --shell fish ./foo/bar/baz | source
- (line 2): function: Illegal function name '__fish_./foo/bar/baz_complete'
function __fish_./foo/bar/baz_complete
^
from sourcing file -
```

Removing the offending Characters only leaving `[a-zA-Z0-9]` or a similar restrictive character-set would result in a workable Fish script.

volkov:
Does completion works after function name fix?

For me (fish 3.1.2) it doesn't.
After changing `complete -c ./foo/bar/baz -f -a '(__fish_foo_bar_baz_complete)'` to `complete -c baz -f -a '(__fish_foo_bar_baz_complete)'` completion starts working as expected.

faho:
>After changing complete -c ./foo/bar/baz -f -a '(__fish_foo_bar_baz_complete)' to complete -c baz -f -a '(__fish_foo_bar_baz_complete)' completion starts working as expected.

https://github.com/fish-shell/fish-shell/issues/6001, fixed in fish 3.3.0.

volkov:
>  fixed in fish 3.3.0.
Nice!


Relative paths should be also supported in generated script body - running completion from different directory would fail now.

[...]

PR #359, opened 2021-11-07
Support of path completion in fish #327

volkov:
First version of absolute path completion for fish. #327 fix

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -290,7 +290,21 @@ By default help string is added as completion description.
 
 You can disable this feature by removing ``_ARGCOMPLETE_DFS`` variable, e.g::
 
-    register-python-argcomplete --shell fish my-awesome-script | grep -v _ARGCOMPLETE_DFS | .
+    register-python-argcomplete --shell fish my-awesome-script | grep -v _ARGCOMPLETE_DFS | source
+
+Absolute Path Completion
+~~~~~~~~~~~~~~~~~~~~~~~~
+If script is not in path you still can register it's completion by specifying absolute path::
+
+    register-python-argcomplete --shell fish /home/awesome-user/my-awesome-script | source
+
+then you can complete it by using ``/home/awesome-user/my-awesome-script`` or ``./my-awesome-script``.
+
+Unfortunately ``~/my-awesome-script`` would not work, to fix it you can run::
+
+    register-python-argcomplete --shell fish my-awesome-script -e /home/awesome-user/my-awesome-script | source
+
+This would enable completion for any ``my-awesome-script`` in any location.
 
 Git Bash Support
 ----------------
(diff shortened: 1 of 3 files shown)


volkov:
@kislyuk @evanunderscore take a look please, I think it could be merged.

evanunderscore:
Looks good to me. Are you able to add a test case to the test suite that verifies this? It could probably go in the base class since I'd expect this to work for Bash too.

README.rst
```diff
@@ -290,7 +290,21 @@ By default help string is added as completion description.
...
+
+    register-python-argcomplete --shell fish /home/awesome-user/my-awesome-script | source
+
+then you can complete it by using ``/home/awesome-user/my-awesome-script`` or ``./my-awesome-script``.
+
+Unfortunately ``~/my-awesome-script`` would not work, to fix it you can run::
+
+    register-python-argcomplete --shell fish my-awesome-script -e /home/awesome-user/my-awesome-script | source
+
+This would enable completion for any ``my-awsome-script`` in any location.
```
kislyuk, line 307:
```suggestion
This would enable completion for any ``my-awesome-script`` in any location.
```

kislyuk (commented):
Thank you @volkov. As @evanunderscore said, this looks good but needs a regression test to make sure the new behavior is covered by the test suite.

kislyuk:
@volkov let us know if you need help adding a test.

volkov:
> Thank you @volkov. As @evanunderscore said, this looks good but needs a regression test to make sure the new behavior is covered by the test suite.

Ok, I will add first version of test to discuss.

volkov:
I have added first version of test https://github.com/kislyuk/argcomplete/pull/359/files#diff-ae070fbd66c55f7bf108994c0a42ffe69eee7319b5bf0f47bef0fcdbd8b496b3R1385

> It could probably go in the base class since I'd expect this to work for Bash too.
I think its better to write separate base test for path completion because `setUp` is different:
* test directory should not be added to `PATH`
* completion should be registered only for path

volkov:
Looks like bash don't resolve relative paths, same test for bash fails on `/prog basic f\t`

evanunderscore:
Is it actually necessary to have a separate class for these tests? Having the test directory on the path shouldn't prevent you from trying to complete an absolute or relative path. If for some reason it is actually necessary to do that, I'd prefer to have the common logic in a base class rather than duplicating it.

I think it makes sense to split the absolute and relative tests into two test functions.

volkov:
> Is it actually necessary to have a separate class for these tests?

Directory on the path is fine, but I can't test path completion when command completion registered.
When completion registered for command `prog` fish would trigger it also for `./prog` and for `test/prog`.
So I can't create test that would pass with both registered completions, but fail with only command completion.

> I'd prefer to have the common logic in a base class rather than duplicating it.
> I think it makes sense to split the absolute and relative tests into two test functions.

I'll try.

volkov:
> So I can't create test that would pass with both registered completions, but fail with only command completion.

Actually it's possible if I have two scripts with different names, e.g `prog` and `frog` - first with command completion, second with path, but I don't want to duplicate test script.

kislyuk:
LGTM, thanks for your patience @volkov.

PR #360, opened 2021-11-07
Fix assertRaisesRegex for Python 2.7

hugovk:
Follow on from https://github.com/kislyuk/argcomplete/pull/357.

Fixes https://github.com/kislyuk/argcomplete/pull/357#issuecomment-962599827.

Maintain support for Python 2.7.


kislyuk:
Thank you for your PR. Python 2.7 is not supported and there will be no further efforts made to maintain existing code that was known to work with it.

kislyuk:
See also: #361

PR #389, opened 2022-11-30
Update importlib-metadata dependency requirements to include versions ^5.0.0

aluuu:
Why:

argcomplete appears to be a dependency of many packages and can't be installed along with packages that require importlib-metadata>=5 pacakge (e.g. [djlint](https://github.com/Riverside-Healthcare/djLint/blob/master/pyproject.toml#L37), in my case, but I'm sure there will be more in future). This change makes such installations possible.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import find_packages, setup
 
 install_requires = []
 tests_require = ["coverage", "flake8", "pexpect", "wheel"]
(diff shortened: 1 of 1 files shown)


aluuu:
Hello @tony @kislyuk,

It doesn't seem that github allows me to request review from you guys via assigning reviewers while creating a PR, so here I am asking for review and the workflow run approval :)

PR #307, opened 2020-06-15
Add option to use tempfiles for IPC

evanunderscore:
Supersedes #306, fixes #267, fixes #142.

Adds support for Git Bash (and potentially other shells on Windows) by giving the option to use a file for IPC instead of file descriptors.

Currently this feature is opt-in as described in the README.

PR #392, opened 2022-12-14
Support PowerShell

bebound:
In https://github.com/kislyuk/argcomplete/pull/307, argcomplete adds the env `ARGCOMPLETE_USE_TEMPFILES` to redirect its output to file and supports Git Bash. We can use the same way to support powershell.

I've used it in https://github.com/Azure/azure-cli/pull/24752, and it works.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -328,6 +328,17 @@ The command line interface of this program must be additionally implemented in a
 
 This option can also be used in combination with the other supported shells.
 
+PowerShell Support
+------------
+To create new completion file, e.g::
+
+    register-python-argcomplete --shell powershell my-awesome-script > ~/my-awesome-script.psm1
+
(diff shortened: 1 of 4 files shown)


IDRozenblad:
Review

kislyuk:
Thanks. How would we go about writing a unit test for this?

bebound:
`pexpect` is not fully compatible with Windows.
I can use `pexpect.popen_spawn.PopenSpawn(f'PowerShell -c "echo 1"',encoding='utf-8')` to get the output of command but I don't know the equivalent of `self.child.sendline`.

bebound:
Sorry, I should have added more dashes at this line. Do I need recreate the whole PR or you can edit this line?
```
PowerShell Support
------------------
```

kislyuk:
Please re-open a new PR (and incorporate your README fix from the newer PR)
as the old one had to be reverted to unblock the release.

PR #398, opened 2023-01-04
Removed part of fish global completion doc after fish-shell fix

volkov:
In fish-shell/fish-shell#8442 global completion was fixed, so we don't need workaround anymore


kislyuk:
Thanks.

PR #405, opened 2023-03-05
Support PowerShell

bebound:
Reopen #392 and fix the README.rst

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@ test_deps:
 
 lint: test_deps
 	flake8 $$(python setup.py --name)
-	for script in scripts/*; do if grep -q python $$script; then flake8 $$script; fi; done
+	for script in scripts/*[^cmd]; do if grep -q python $$script; then flake8 $$script; fi; done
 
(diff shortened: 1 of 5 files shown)


bebound:
There are still some errors, I'll fix them.

kislyuk:
Thank you

kislyuk:
@bebound can you please update this so that the file `register-python-argcomplete.cmd` does not get installed on non-Windows platforms by default?

kislyuk:
I have deleted the file `scripts/register-python-argcomplete.cmd` and will not be adding it back. In general, scripts for contributor-supported shells will need to be installed by separate means activated via setuptools extras.

bebound:
Windows user cannot run `register-python-argcomplete` without this cmd file.

kislyuk:
I understand. We need to find a way to only install this file on Windows.

bebound:
How about using `entry_points`? https://github.com/bebound/argcomplete/tree/entry-point
It works on Linux and Windows. I don't know if it is compatible with tch.

PR #406, opened 2023-03-05
setup.py: exclude test.* subpackages from find_packages too

thesamesam:
Otherwise, test.test_package will be picked up when installing via PEP517.

Bug: https://bugs.gentoo.org/899628


kislyuk:
Thank you

thesamesam:
Thanks!

PR #410, opened 2023-03-16
Make fish optional.

mcepl:
The test suite should not fail because of missing fish on the system.

We are maintaining this package for SLE, where fish is not present (or very old one), and I don’t think it is necessary for building the package to make fish actually required. If it is not present, then the fish-related functionality won’t be tested and that’s it.


kislyuk (approved):

kislyuk:
Thanks. Agreed, fish is optional.

kislyuk:
Thanks for the PR. I ended up going with a slightly different approach. Fish tests are now disabled by default.

PR #422, opened 2023-04-07
Note completers can return iterables of strings, not just lists

scop:
Some of the built in implementations actually do that already.


kislyuk:
Thanks! And thanks for all your work on bash-completion.

Issue #420, opened 2023-03-26
test_console_script_module_wheel and test_console_script_package_wheel fail on Python 3.7 with importlib_metadata >= 5


kislyuk:
Unclear if this is an actual issue or a test case artifact.
```
======================================================================
FAIL: test_console_script_module_wheel (__main__.TestBashGlobal)
Test completing a console_script for a module from a wheel.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./test/test.py", line 1326, in test_console_script_module_wheel
    self._test_console_script(wheel=True)
  File "./test/test.py", line 1311, in _test_console_script
    self.assertEqual(self.sh.run_command(command), "arg\r\n")
AssertionError: "usage: test-module [-h] {arg}\r\ntest-mo[66 chars]\r\n" != 'arg\r\n'
+ arg
- usage: test-module [-h] {arg}
- test-module: error: argument arg: invalid choice: 'a' (choose from 'arg')
```
```
======================================================================
FAIL: test_console_script_package_wheel (__main__.TestBashGlobal)
Test completing a console_script for a package from a wheel.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "./test/test.py", line 1331, in test_console_script_package_wheel
    self._test_console_script(package=True, wheel=True)
  File "./test/test.py", line 1311, in _test_console_script
    self.assertEqual(self.sh.run_command(command), "arg\r\n")
AssertionError: "usage: test-package [-h] {arg}\r\ntest-p[68 chars]\r\n" != 'arg\r\n'
+ arg
- usage: test-package [-h] {arg}
- test-package: error: argument arg: invalid choice: 'a' (choose from 'arg')


----------------------------------------------------------------------
```

flu0r1ne:
I've recently noticed an issue and it's highly likely that it's related to #436.

hroncok:
I get the same failures in Fedora with Python 3.12.0b2 even with(out) https://github.com/kislyuk/argcomplete/pull/436:

```
=================================== FAILURES ===================================
__________________ TestBashGlobal.test_console_script_module ___________________

self = <test.test.TestBashGlobal testMethod=test_console_script_module>

    @unittest.skipIf(os.uname()[0] == "Darwin", "Skip test that fails on MacOS")
    def test_console_script_module(self):
        """Test completing a console_script for a module."""
>       self._test_console_script()

test/test.py:1293: 
[...]

PR #424, opened 2023-04-21
allow importlib-metadata 6.x

reaperhulk:
I'm not actually sure if this works, but the 3.7 CI jobs should hopefully show whether this is safe 😄

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from setuptools import find_packages, setup
 
 install_requires = []
 tests_require = ["coverage", "pexpect", "wheel", "ruff", "mypy"]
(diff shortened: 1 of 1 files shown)


kislyuk:
Thanks. I'm seeing an issue in Python 3.7 CI, but I'm not certain that it's related to importlib-metadata. Need to investigate.

reaperhulk:
Yeah, it appears there's a failure that also occurs in 5.x (as you filed an issue about it in #420)

kislyuk:
OK, Python 3.7 is EOL soon anyway so I've marked these tests as skip as well.

Released in v3.0.6. If anyone sees any issues with importlib_metadata > 4 on Python 3.7, please let me know and we can troubleshoot.

Issue #363, opened 2021-11-08
Migrate to pyproject.toml and flit


kislyuk:

tony:
@kislyuk This is somewhat separate:

- Do you want setup.py converted to pyproject.toml ([PEP 518](https://peps.python.org/pep-0518/), [PEP 621](https://peps.python.org/pep-0621/), example https://github.com/thisch/pytest-sphinx/pull/41/files))
- More heavy duty:

  - Would you like to try [hatch](https://hatch.pypa.io/latest/) (https://github.com/pypa/hatch)? Maintained by PyPA, has a `hatch shell` command? I heard about it today and it looks promising. I can try it in a separate PR if you'd like

    P.S. I haven't tried it before actually, I use poetry 1.2 on all my projects (and it's very mature at this stage)

kislyuk:
Thanks for the offer. I'm open to a PR for migrating setup.py contents to pyproject.toml, but not to hatch.

halfow:
Hi @kislyuk !

This looked like the ball was dropped on so I picked it up :)

I took a look at using the regular `setuptools` for migration to pure `pyproject.toml` as it's the one already used.  
Here are things that I noted looking at a possible migration: 

1. `pyproject.toml` `setuptools`  is still in beta. *(guess this is because of all the legacy it has)*
2. `setup.py` `scripts` has been deprecated and `project.scripts` in  `pyproject.toml` is recommended as the migration path *(`entry_points` instead of stand alone scripts)*
3. `setup.py` `url` is deprecated in `pyproject.toml` and use of `project.urls` seems to be encouraged. This one has impact on `pip show` as Home-page will be missing *(Beta things I guess)*. 
4. Are the `setup.cfg` `bdist_rpm` used? Latest RPM release I find is 1.8.1. This makes me believe that this is not used anymore. 


Thanks for this awesome project!

kislyuk:
Hi @halfow , I left a review in the PR. To your other questions-
- We will not be setting any setuptools build settings or any other build tool settings in pyproject.toml for now. The beta restriction is very apt - in my experience, storing build settings in pyproject.toml is just not stable enough to be reliable for now.
- We will not be moving away from setuptools scripts for now
- We will not be removing any currently present fields marked as "deprecated" for now
- I'm not certain whether bdist_rpm is actively being used, but I'm not ready to remove it at this time.

halfow:
Thank you for the clarification and answers :) 

I will update accordingly.

PR #427, opened 2023-05-14
setup.py -> pyproject.toml migration start

halfow:
#363 

A starting point to make it easier to talk about

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,28 +1,9 @@
-# -------------------------------------------------------------------------------------------------
-# Build system settings
-# -------------------------------------------------------------------------------------------------
 [build-system]
 requires = ["setuptools>=67.7.2", "wheel"]
 build-backend = "setuptools.build_meta"
 
-[tool.setuptools]
-include-package-data = true
-platforms = ["MacOS X", "Posix"]
-zip-safe = false
-
-[tool.setuptools.packages.find]
-include = ["argcomplete"]
-namespaces = false # to disable scanning PEP 420 namespaces (true by default)
-
-[tool.setuptools.package-data]
-argcomplete = ["py.typed", "bash_completion.d/python-argcomplete"]
-
-# -------------------------------------------------------------------------------------------------
-# Project Metadata
-# -------------------------------------------------------------------------------------------------
 [project]
 name = "argcomplete"
-version = "3.0.8"
 description = "Bash tab completion for argparse"
 readme = "README.rst"
 requires-python = ">=3.6"
@@ -61,10 +42,6 @@ test = ["coverage", "pexpect", "wheel", "ruff", "mypy"]
 "Issue Tracker"= "https://github.com/kislyuk/argcomplete/issues"
 "Change Log"= "https://github.com/kislyuk/argcomplete/blob/master/Changes.rst"
 
-
-# -------------------------------------------------------------------------------------------------
-# Linter and formatter settings
-# -------------------------------------------------------------------------------------------------
 [tool.black]
 line-length = 120
 exclude = ".*/version.py"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -1,3 +1,70 @@
...
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+  "Development Status :: 5 - Production/Stable",
+  "Topic :: Software Development",
+  "Topic :: Software Development :: Libraries :: Python Modules",
+  "Topic :: System :: Shells",
+  "Topic :: Terminals",
+]
+dependencies = ["importlib-metadata >= 0.23, < 7; python_version < '3.8'"]
```
kislyuk, line 53: Please update this to be conditional on the Python version, as in the original.

kislyuk (changes requested):
Hi, thanks for your PR. Frankly I don't consider a total migration to pyproject.toml to be a priority, because I don't think the pyproject.toml based build tooling ecosystem is sufficiently mature to be reliable. We won't be moving away from setuptools scripts for now either.

I'm happy to move the "project metadata" settings to pyproject.toml, but I won't be setting any build tool settings in pyproject.toml.

If you want to proceed with this PR, please address the review comments and we can merge. The biggest lift is actually the version auto-updating issue.

pyproject.toml
halfow: Did see the `python_version < '3.8'` on the end or do you want it more explicit for the versions?
_`python_version < '3.8'` + the `python_requires=">=3.6"` would give the span all of 3.6 and 3.7_ 

More explicit, could be something like: 
````
dependencies = [
    "importlib-metadata >= 0.23, < 7; python_version == '3.6'",
    "importlib-metadata >= 0.23, < 7; python_version == '3.7'",
] 
````

File: pyproject.toml [hunk 2]
```diff
@@ -1,3 +1,70 @@
...
+include-package-data = true
+platforms = ["MacOS X", "Posix"]
+zip-safe = false
+
+[tool.setuptools.packages.find]
+include = ["argcomplete"]
+namespaces = false # to disable scanning PEP 420 namespaces (true by default)
+
+[tool.setuptools.package-data]
+argcomplete = ["py.typed", "bash_completion.d/python-argcomplete"]
```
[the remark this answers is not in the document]
halfow: Would be nice to keep `build-system` [PEP-518](https://peps.python.org/pep-0518/)

pyproject.toml
kislyuk: Never mind, I misread the original line, thanks.

kislyuk:
Thanks for making the changes!

It looks like setuptools wants version to be in pyproject.toml if `project` is in pyproject.toml. I think this forces a migration of my release tooling (common.mk) to setuptools_scm.

halfow:
@kislyuk how do you want to proceed? 

I can make the tweaks to use `setuptool-scm` for versioning if that is something that is in the backlog and should be done any way.  :)

kislyuk:
@halfow would love it if you could contribute that. You'd need to edit common.mk to update the release scripting. Here is another project of mine that already uses setuptools_scm, for reference: https://github.com/pyauth/tsp-client

pyproject.toml
```diff
@@ -42,6 +43,8 @@ test = ["coverage", "pexpect", "wheel", "ruff", "mypy"]
 "Issue Tracker"= "https://github.com/kislyuk/argcomplete/issues"
 "Change Log"= "https://github.com/kislyuk/argcomplete/blob/master/Changes.rst"
 
+[tool.setuptools_scm]
```
halfow, line 46: Described way from https://pypi.org/project/setuptools-scm/ .
Would this be alright @kislyuk or should I see if it can be moved to to setup.py?

kislyuk: This is fine.

kislyuk (approved):

kislyuk:
Thanks.

PR #434, opened 2023-05-29
Use ` as escape character in PowerShell

bebound:
Related issue: https://github.com/Azure/azure-cli/issues/26526

Argcomplete escape the special_chars with `\` when get completion.
The result is incorrect on PowerShell.
For example. space will be converted to `\ `, but it should be `` `  ``

### DIFF
--- a/argcomplete/finders.py
+++ b/argcomplete/finders.py
@@ -520,7 +520,11 @@ class CompletionFinder(object):
             completions = [c.replace("'", r"'\''") for c in completions]
 
         # PowerShell uses ` as escape character.
-        escape_char = '`' if os.getenv('PSModulePath') else "\\"
+        if os.getenv('PSModulePath'):
+            escape_char = '`'
+            special_chars = special_chars.replace('`', '')
+        else:
+            escape_char = "\\"
         for char in special_chars:
             completions = [c.replace(char, escape_char + char) for c in completions]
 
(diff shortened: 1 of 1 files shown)


argcomplete/finders.py
```diff
@@ -519,8 +519,14 @@ def quote_completions(
             special_chars = ""
             completions = [c.replace("'", r"'\''") for c in completions]
 
+        # PowerShell uses ` as escape character.
+        if os.getenv('PSModulePath'):
+            escape_char = '`'
+            special_chars = special_chars.replace('`', '')
```
bebound, line 525: On Linux, why `\` is not in special chars?

Do I need to convert `` ` `` to some other string, and convert back to ``` `` ```?

kislyuk:
Thanks. It would be best to have a unit test for this, but we've never managed to get powershell to run in unit tests. I trust you have tested this and it works for you.

kislyuk:
Actually, I think there is a problem that will have to be corrected before I release this. I don't think the presence of `PSModulePath` is a sufficiently positive signal that we're running in powershell. It might be set in other shells through some sort of common configuration mechanism.

Instead, the variable `_ARGCOMPLETE_SHELL` should be set to `powershell` in https://github.com/kislyuk/argcomplete/blob/develop/argcomplete/shell_integration.py#L101-L107 and its value matched instead of `PSModulePath`.

kislyuk:
@bebound I fixed the issue with the environment variables. Will you please confirm that the fix actually worked for you and report the version of powershell that you used?

bebound:
It works on PS 5.1.22621.1778, thanks.

PR #461, opened 2023-11-21
enhance scripts/activate-global-python-argcomplete

zoumingzhe:
I'm developing tab completion based on argcomplete, and activate-global-python-argcomplete sometimes requires yes confirmation, preventing me from fully automating.


zoumingzhe:
@kislyuk Can you help review it?

kislyuk:
LGTM, thanks (although the try..except clause does not make any sense, I will fix it separately).

Issue #462, opened 2023-11-23
How do I exclude files in the users directory from the tab completion?


aidan-gallagher:
For my tool there is no reason a user should pass a local file to the script. Therefore I don't want the files to appear in the tab completion. 

Here is an example of the output I see. On the last command (tab completed) I don't want files to appear just the argparse arguments.

```
aidan@ag474u:~/Code/Per/dpkg-buildenv$ ls
built_packages  debian  developer-packages.txt  dpkg-buildenv  Jenkinsfile  README.md  tmp  workspaces




aidan@ag474u:~/Code/Per/dpkg-buildenv$ dpkg-buildenv/dpkg_buildenv.py --help
usage: dpkg_buildenv.py [-h] [-nc] [-s SOURCES] [-it] [-d] [command]

positional arguments:
  command               Command to execute in the container. [Docker run argument]

optional arguments:
  -h, --help            show this help message and exit
  -nc, --no-cache       Do not use cache when building the image. [Docker build argument]
  -s SOURCES, --sources SOURCES
                        Select a sources file stored at /etc/dpkg-buildenv/sources.list.d/<SOURCE>.list.
  -it, --interactive-tty
                        Open an interactive terminal to the container. [Docker run argument]
  -d, --delete-images   Delete all build environment images generated by this tool




aidan@ag474u:~/Code/Per/dpkg-buildenv$ dpkg-buildenv/dpkg_buildenv.py  (<----This is where I hit tab)
built_packages/         --delete-images         .git/                   --help                  Jenkinsfile             .pytest_cache/          --sources
-d                      developer-packages.txt  .gitignore              --interactive-tty       -nc                     README.md               tmp/
debian/                 dpkg-buildenv/          -h                      -it                     --no-cache              -s                      workspaces/
aidan@ag474u:~/Code/Per/dpkg-buildenv$ dpkg-buildenv/dpkg_buildenv.py 
```

evanunderscore:
Are you using zsh? This may be a duplicate of #458.

evanunderscore:
Fixed in #463.

Issue #458, opened 2023-11-13
installing zsh globally adds paths to completion options introduced in v3.1.4


maurerle:
When installing `argcomplete >= v3.1.4` globally, the default folder and file options are added (additionally to the useful completions).

```bash
(myenv) ➜  MYPROJECT git:(main) ✗ project --loglevel
myproject/        docker_configs/           LICENSES/               
logfile           Dockerfile                poetry.lock             
cli.py            docs/                     __pycache__/            
dist/             export/                                           
CRITICAL  WARNING  ERROR                                                    
NOTSET    FATAL    DEBUG  WARN  INFO  -- logging level used for file log 
```

#### Older Version and non-global works fine

pip install argcomplete==v3.1.3` and installing the tab completion, everything works fine.

So the fix here is probably missing something:
https://github.com/kislyuk/argcomplete/commit/afdf75bc6f43f1d1584468665ea1147c22b9767c

evanunderscore:
Fixed in #463. Thanks for reporting!

Issue #454, opened 2023-10-30
Messes with default zsh completion for redirects.


jmcantrell:
I recently installed pipx and this was a required dependency. Some time after that, I discovered that zsh completion after the redirect operator was not working. After some assistance from #zsh, it  seems like argcomplete may be the culprit. 

Here's the completion trace after trying to complete an empty redirection (`>./<tab>`) from zsh (with no configuration, and just compinit run):

```
+_complete_debug:24> : zsh 5.9
+_complete_debug:25> _main_complete
+_main_complete:11> local IFS=$' \t\n\C-@'
+_main_complete:25> eval $'local -A _comp_caller_options;\n             _comp_caller_options=(${(kv)options[@]});\n             setopt localoptions localtraps localpatterns ${_comp_options[@]};\n             local IFS=$\' \\t\\r\\n\\0\';\n             builtin enable -p \\| \\~ \\( \\? \\* \\[ \\< \\^ \\# 2>&-;\n             exec </dev/null;\n             trap - ZERR;\n             local -a reply;\n             local REPLY;\n             local REPORTTIME;\n             unset REPORTTIME'
 +(eval):1> local -A _comp_caller_options
[...]

PR #433, opened 2023-05-28
Fix ZSH implementation - missing completion

flu0r1ne:
The ZSH auto-completion was malfunctioning on a fresh Ubuntu 22.04 installation. This issue originated from several problems within the global auto-completion script.

Initially, `COMP_WORDS` was employed for matching modules and interpreter arguments. However, `COMP_WORDS` isn't defined in ZSH; `words` is the correct variable to use instead. To resolve this, an argument vector with base-1 indexing was created to work with both shells.

Secondly, the __python_argcomplete_scan_head function included a defective read command in ZSH during interpreter matching. Using -k with read in ZSH reads a specified num of bytes, regardless of a newline presence. This issue was addressed by introducing a secondary read command in both shells to retrieve the line. Additionally, in ZSH, regex matches are not stored in BASH_REMATCH unless a compatibility option is activated. This was addressed by enabling this option locally. The difference between base-0 and base-1 indexing was mitigated by incrementing `BASH_REMATCH` when operating in Bash and indexing correctly.

A ZSH-compatible version of __python_argcomplete_expand_tilde_by_ref was also introduced as the prior version utilized zero indexing and Bash-specific indirection syntax.

Moreover, the indexing and variable expansion of the argument array in the `__python_argcomplete_run` function were correctly adjusted.

Lastly, the path to a Python script was expanded using tilde notation to enable auto-completion when employing:

```zsh
python ~/some/script.py
```

TESTED WITH:

GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu) zsh 5.8.1 (x86_64-ubuntu-linux-gnu)
Linux ubuntu-22-04-2-desktop 5.19.0-42-generic

PR #463, opened 2023-11-25
Fix and test global completion in zsh

### LINKED ISSUE #451: zsh: Extra space is added during hierarchy completion (directory or custom)
Hello,
Please find example in attachment.
[test_comp.py.zip](https://github.com/kislyuk/argcomplete/files/12866822/test_comp.py.zip)

When I try to complete on zsh:
./test_comp.py --dir /ho\<tab\>
generates: "/home/ " with extra space. As a result - all child items are not available 

./test_comp.py --custom /pa\<tab\>
generates: "/parent/ " also with extra space

When I try to complete on bash - no issues are found.

Environment:
lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 22.04.3 LTS
Release:        22.04
Codename:       jammy

pip show argcomplete
Name: argcomplete
Version: 3.1.2
Summary: Bash tab completion for argparse
Home-page: https://github.com/kislyuk/argcomplete
Author: Andrey Kislyuk
Author-email: kislyuk@gmail.com
License: Apache Software License
Location: /home/dybala/.local/lib/python3.10/site-packages
Requires: 
Required-by: 

python --version
Python 3.10.12

zsh --version
zsh 5.8.1 (x86_64-ubuntu-linux-gnu)

bash --version
GNU bash, version 5.1.16(1)-release (x86_64-pc-linux-gnu)

evanunderscore:
Fixes #451, #458, #462

### DIFF
--- a/argcomplete/bash_completion.d/_python-argcomplete
+++ b/argcomplete/bash_completion.d/_python-argcomplete
@@ -138,9 +138,6 @@ _python_argcomplete_global() {
         req_argv=( "" "${COMP_WORDS[@]:1}" )
         __python_argcomplete_expand_tilde_by_ref executable
     else
-        if [[ "$service" != "-default-" ]]; then
-            return
-        fi
         executable="${words[1]}"
         req_argv=( "${words[@]:1}" )
     fi
@@ -240,5 +237,8 @@ if [[ -z "${ZSH_VERSION-}" ]]; then
     complete -o default -o bashdefault -D -F _python_argcomplete_global
 else
     autoload is-at-least
-    compdef _python_argcomplete_global -P '*'
+    # Replace only the default completer (_default).
+    # There are many other special contexts we don't want to override.
+    # https://zsh.sourceforge.io/Doc/Release/Completion-System.html
+    compdef _python_argcomplete_global -default-
 fi
--- a/argcomplete/bash_completion.d/_python-argcomplete
+++ b/argcomplete/bash_completion.d/_python-argcomplete
@@ -139,9 +139,6 @@ _python_argcomplete_global() {
         __python_argcomplete_expand_tilde_by_ref executable
     else
         if [[ "$service" != "-default-" ]]; then
-            # TODO: this may not be sufficient - see https://zsh.sourceforge.io/Doc/Release/Completion-System.html
-            # May need to call _complete with avoid-completer=_python-argcomplete or something like that
-            _default
             return
         fi
         executable="${words[1]}"
@@ -208,7 +205,15 @@ _python_argcomplete_global() {
                 _ARGCOMPLETE_SHELL="zsh" \
                 _ARGCOMPLETE_SUPPRESS_SPACE=1 \
                 __python_argcomplete_run "$executable" "${(@)req_argv[1, ${ARGCOMPLETE}-1]}"))
-            _describe "$executable" completions
+            local nosort=()
+            local nospace=()
+            if is-at-least 5.8; then
+                nosort=(-o nosort)
+            fi
+            if [[ "${completions-}" =~ ([^\\]): && "${BASH_REMATCH[2]}" =~ [=/:] ]]; then
+                nospace=(-S '')
+            fi
+            _describe "$executable" completions "${nosort[@]}" "${nospace[@]}"
         else
             COMPREPLY=($(IFS="$IFS" \
                 COMP_LINE="$COMP_LINE" \
@@ -234,5 +239,6 @@ _python_argcomplete_global() {
 if [[ -z "${ZSH_VERSION-}" ]]; then
     complete -o default -o bashdefault -D -F _python_argcomplete_global
 else
+    autoload is-at-least
     compdef _python_argcomplete_global -P '*'
 fi
--- a/argcomplete/shell_integration.py
+++ b/argcomplete/shell_integration.py
@@ -42,7 +42,15 @@ _python_argcomplete%(function_suffix)s() {
             _ARGCOMPLETE_SHELL="zsh" \
             _ARGCOMPLETE_SUPPRESS_SPACE=1 \
             __python_argcomplete_run ${script:-${words[1]}}))
-        _describe "${words[1]}" completions -o nosort
+        local nosort=()
+        local nospace=()
+        if is-at-least 5.8; then
+            nosort=(-o nosort)
+        fi
+        if [[ "${completions-}" =~ ([^\\]): && "${match[1]}" =~ [=/:] ]]; then
+            nospace=(-S '')
+        fi
+        _describe "${words[1]}" completions "${nosort[@]}" "${nospace[@]}"
     else
         local SUPPRESS_SPACE=0
         if compopt +o nospace 2> /dev/null; then
@@ -67,6 +75,7 @@ _python_argcomplete%(function_suffix)s() {
 if [[ -z "${ZSH_VERSION-}" ]]; then
     complete %(complete_opts)s -F _python_argcomplete%(function_suffix)s %(executables)s
 else
+    autoload is-at-least
     compdef _python_argcomplete%(function_suffix)s %(executables)s
 fi
 """
--- a/test/test.py
+++ b/test/test.py
@@ -1321,6 +1321,15 @@ class TestZsh(TestBashZshBase, unittest.TestCase):
 class TestBashZshGlobalBase(TestBashZshBase):
     install_cmd = 'eval "$(activate-global-python-argcomplete --dest=-)"'
 
+    def test_redirection_completion(self):
+        with TempDir(prefix="test_dir_py", dir="."):
+            self.sh.run_command("cd " + os.getcwd())
+            self.sh.run_command("echo failure > ./foo.txt")
+            self.sh.run_command("echo success > ./foo.\t")
+            with open("foo.txt") as f:
+                msg = f.read()
+            self.assertEqual(msg, "success\n")
+
     def test_python_completion(self):
         self.sh.run_command("cd " + TEST_DIR)
         self.assertEqual(self.sh.run_command("python3 ./prog basic f\t"), "foo\r\n")
--- a/test/test.py
+++ b/test/test.py
@@ -1369,9 +1369,6 @@ class TestBashZshGlobalBase(TestBashZshBase):
             command = "pip install {} --target .".format(test_package)
             if not wheel:
                 command += " --no-binary :all:"
-                if sys.platform == "darwin":
-                    # Work around https://stackoverflow.com/questions/24257803
-                    command += ' --install-option="--prefix="'
             install_output = self.sh.run_command(command)
             self.assertEqual(self.sh.run_command("echo $?"), "0\r\n", install_output)
             command = "test-module"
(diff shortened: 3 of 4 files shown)


argcomplete/bash_completion.d/_python-argcomplete
```diff
@@ -139,9 +139,6 @@ _python_argcomplete_global() {
         __python_argcomplete_expand_tilde_by_ref executable
     else
         if [[ "$service" != "-default-" ]]; then
-            # TODO: this may not be sufficient - see https://zsh.sourceforge.io/Doc/Release/Completion-System.html
-            # May need to call _complete with avoid-completer=_python-argcomplete or something like that
-            _default
```
evanunderscore, line 144: @kislyuk This was added in https://github.com/kislyuk/argcomplete/commit/afdf75bc6f43f1d1584468665ea1147c22b9767c, but I'm not sure what the intent was. This was causing default completions to be mixed in with the returned completions in my tests, and removing it doesn't seem to have caused any of the existing tests to fail. If you remember what problem this was addressing, I can take another look into it.

argcomplete/bash_completion.d/_python-argcomplete
```diff
@@ -208,7 +205,15 @@ _python_argcomplete_global() {
                 _ARGCOMPLETE_SHELL="zsh" \
                 _ARGCOMPLETE_SUPPRESS_SPACE=1 \
                 __python_argcomplete_run "$executable" "${(@)req_argv[1, ${ARGCOMPLETE}-1]}"))
-            _describe "$executable" completions
+            local nosort=()
+            local nospace=()
+            if is-at-least 5.8; then
+                nosort=(-o nosort)
+            fi
+            if [[ "${completions-}" =~ ([^\\]): && "${BASH_REMATCH[2]}" =~ [=/:] ]]; then
```
evanunderscore, line 213: Note this must use `BASH_REMATCH` (as opposed to `match`) because this script uses `setopt local_options BASH_REMATCH`. The indexing is different because we haven't also set `KSH_ARRAYS` and I haven't called `__python_argcomplete_upshift_bash_rematch`.

argcomplete/bash_completion.d/_python-argcomplete
```diff
@@ -208,7 +205,15 @@ _python_argcomplete_global() {
                 _ARGCOMPLETE_SHELL="zsh" \
                 _ARGCOMPLETE_SUPPRESS_SPACE=1 \
                 __python_argcomplete_run "$executable" "${(@)req_argv[1, ${ARGCOMPLETE}-1]}"))
-            _describe "$executable" completions
+            local nosort=()
+            local nospace=()
+            if is-at-least 5.8; then
+                nosort=(-o nosort)
+            fi
```
evanunderscore, line 212: The `-o` argument was removed in #433 due to incompatibility with zsh 5.7. The arguments to `-o` were added in 5.8:

> Changes between 5.7 and 5.8
> [...]
> The compadd builtin's -o option now takes an optional argument to specify the order of completion matches. This affects the display of candidate matches and the order in which they are selected when cycling between them using menu completion.

https://zsh.sourceforge.io/releases.html

argcomplete/shell_integration.py
```diff
@@ -42,7 +42,15 @@
             _ARGCOMPLETE_SHELL="zsh" \
             _ARGCOMPLETE_SUPPRESS_SPACE=1 \
             __python_argcomplete_run ${script:-${words[1]}}))
-        _describe "${words[1]}" completions -o nosort
+        local nosort=()
+        local nospace=()
+        if is-at-least 5.8; then
+            nosort=(-o nosort)
+        fi
+        if [[ "${completions-}" =~ ([^\\]): && "${match[1]}" =~ [=/:] ]]; then
```
evanunderscore, line 50: Note the different indexing semantics between `match` and `BASH_REMATCH`. Both versions are extracting the first match.

test/test.py
```diff
@@ -1248,9 +1248,6 @@ def setUp(self):
         path = ":".join([os.path.join(BASE_DIR, "scripts"), TEST_DIR, "$PATH"])
         sh.run_command("export PATH={0}".format(path))
         sh.run_command("export PYTHONPATH={0}".format(BASE_DIR))
-        if self.repl_provider == bash_repl:
-            # Disable the "python" module provided by bash-completion
-            sh.run_command("complete -r python python2 python3")
```
evanunderscore, line 1253: This was not actually doing anything, since `self.repl_provider` is a separate method that calls `bash_repl`. Since the tests were still passing, I'm assuming this is no longer needed.

argcomplete/shell_integration.py
evanunderscore: The returned completion format is `completion:description`, where colons are escaped with backslashes. This looks for the character before the first unescaped colon and checks it against the same characters used for bash.

argcomplete/bash_completion.d/_python-argcomplete
kislyuk: @evanunderscore please do not delete this without a replacement, and especially please do not delete the comment. This was added in response to https://github.com/kislyuk/argcomplete/issues/454. Specifically, it was found that the default completer interferes with zsh completion in the redirect target context (words after `>` or `2>`).

I didn't manage to write a test for this, and I since realized that this was a stopgap solution that will need to be further refined. For example, instead of calling _default unconditionally as a fallback here, we may need to call it only when certain contexts like `-redirect-` are listed in `$service` (see the link in the comment for a full list of contexts - there are others in the list that will almost certainly need similar treatment).

evanunderscore: Thanks for the issue link, I did search initially but wasn't able to find it.

I've pushed a fix in 1f6e66148ae8a4225481b4f3a9ae054d462cede0. Rather than handle all of the special contexts, I've updated the `compdef` command to only override `-default-` which leaves the other special contexts undisturbed. The tests are all passing and I've poked around manually in the shell for a bit, and it seems to be working as expected.

Are you happy with this fix and the comment I've left?

evanunderscore: Unfortunately the macos tests are failing and I don't have a good way to troubleshoot them. I can just disable the global zsh tests on macos (since they weren't running at all before I added them), but it is strange that they were all passing before I made the `compdef` change.

evanunderscore: The issue is that some of the runners already have `python-argcomplete` installed as a dependency of `pipx`, which explains why the failures seemed completely inconsistent - the tests were actually testing whichever version happened to be installed and not what was in the source. For now, I've added a step to uninstall it, but a better solution would be to revisit the change I suggested in #255. Without it, there is no good way to avoid this mismatch.

test/test.py
```diff
@@ -1375,27 +1377,32 @@ def _test_console_script(self, package=False, wheel=False):
             command += " a\t"
             self.assertEqual(self.sh.run_command(command), "arg\r\n")
 
-    @unittest.skipIf(os.uname()[0] == "Darwin", "Skip test that fails on MacOS")
```
evanunderscore, line 1378: These tests all pass in CI now, but I vaguely remember you were having trouble running them at some point. @kislyuk can you please pull this branch and confirm that the tests pass for you locally?

kislyuk: Confirming that all tests pass on this branch for me.

argcomplete/bash_completion.d/_python-argcomplete
kislyuk: > Rather than handle all of the special contexts, I've updated the compdef command to only override -default- which leaves the other special contexts undisturbed.

Sounds good, thanks.

kislyuk:
Thanks @evanunderscore. I'll do some more testing with the new compdef change and then roll a release.

kislyuk:
@evanunderscore after this change, global completion doesn't work in zsh for me, on both Linux and macos. How did you test it?

kislyuk:
@evanunderscore I have reverted the part of this PR that changes the zsh compdef line (and re-added the call to _default).

It doesn't look like `compdef -default-` will work for global completion.

kislyuk:
Actually, since the revert of just that part breaks tests, I'm going to go ahead and revert the whole PR; please reopen it when you have a chance and see if your changes can be reconciled with a version of the compdef line that works.

kislyuk:
OK, I think I found a way to preserve the behavior that we want - I've replaced `-default-` with `-first` in the compdef line (but note the autoload definition line, the first line of the file, remains `#compdef -P *`, which is also key). However the tests are now failing, can you please take a look?

In light of the apparent fragility of this change, it will require further testing and hopefully we can find a way to exercise it in CI. Note that it can be tricky to make sure that zsh unloaded the previous function and loaded the one that you edited - I had to remove `~/.zcompdump` sometimes to make the changes effective.

evanunderscore:
@kislyuk The tests are failing because you've reintroduced the problem from #458 by not suppressing the default completions. As an example, `test_wordbreak_chars` is expecting this:

```
% prog break a<tab>
% prog break a:b:<user types c>
```

The completion of `:b:` is unambiguous between the two possible completions, so is inserted when the user presses tab. However if default completion is not suppressed, there is a third option:

```
% prog break a<tab>
Completing test/prog
a:b:d  a:b:c  --                                                              
Completing file
argcomplete/
% prog break a<user types c>
```

Now it doesn't insert anything, and the resulting line is just `ac`, hence the error message.

I agree it's not ideal for unrelated tests to fail like this. The fix would be to have the shell tests `cd` into an empty directory so that filenames can't incidentally ruin tests, but we would have to add another test that intentionally tries to trigger this problem so we don't end up with another regression for #458.

> @evanunderscore after this change, global completion doesn't work in zsh for me, on both Linux and macos. How did you test it?

I tested on Ubuntu 22.04 with zsh 5.8.1 using the same method as the tests use, i.e. `'eval "$(activate-global-python-argcomplete --dest=-)"'`. How did you test it? What happened when you tested it, e.g. did completion break entirely or was it only argcomplete completions that did not work? Did you get any useful output when you set `_ARC_DEBUG=1`?

> In light of the apparent fragility of this change, it will require further testing and hopefully we can find a way to exercise it in CI. Note that it can be tricky to make sure that zsh unloaded the previous function and loaded the one that you edited - I had to remove ~/.zcompdump sometimes to make the changes effective.

To the best of my understanding, the change *was* being exercised in CI. The problem will have to be either with a mismatch between how you and I tested locally, or the version of zsh you're using, or something else specific to your environment.

evanunderscore:
> OK, I think I found a way to preserve the behavior that we want - I've replaced -default- with -first in the compdef line (but note the autoload definition line, the first line of the file, remains #compdef -P *, which is also key). However the tests are now failing, can you please take a look?

The `#compdef -P *` is the part that I missed - I didn't realise there was any significance to it (and I don't think there is when the file is run directly). For reference, it's documented under "20.2.2 Autoloaded files" here: https://zsh.sourceforge.io/Doc/Release/Completion-System.html

Changing this to `#compdef -default-` to match the compdef line seems to fix it for me when installing globally. I'm not totally sure I understand how that comment and the actual `compdef` command interact, or if we're using them correctly.

Can you try reverting your fix and changing the #compdef comment please?

PR #466, opened 2023-12-08
Fix zsh global install

evanunderscore:
This fixes zsh global completion breakage introduced in #463, and updates the test to use a similar method as `activate-global-python-argcomplete` for zsh.

The zsh global completion script was previously mixing paradigms; it seems that you are supposed to use a file with a `#compdef` comment which is autoloaded, **or** manually call `compdef` yourself. I've updated this to only use the `#compdef` comment, but it does feel like everything would be simpler if we went the other way and sourced the file directly as we do with bash. The only catch with that is we would probably have to append this to `~/.zshrc` instead of `~/.zshenv` since we need `compdef` to be available, and it seems undesirable to `autoload compdef` on the user's behalf.

### DIFF
--- a/argcomplete/bash_completion.d/_python-argcomplete
+++ b/argcomplete/bash_completion.d/_python-argcomplete
@@ -1,9 +1,14 @@
-#compdef -P *
+#compdef -default-
 
 # Copyright 2012-2023, Andrey Kislyuk and argcomplete contributors.
 # Licensed under the Apache License. See https://github.com/kislyuk/argcomplete for more info.
 
 # Note: both the leading underscore in the name of this file and the first line (compdef) are required by zsh
+
+# In zsh, this file is autoloaded and used as the default completer (_default).
+# There are many other special contexts we don't want to override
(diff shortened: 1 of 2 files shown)


kislyuk:
@evanunderscore as far as I know, we can't source the file directly (at least not without making our installation process a lot less general). Zsh has no default facility to source completion scripts at startup - instead it scans the first lines of these files for compdef declarations. The actual scripts are sourced on demand according to those declarations.

By the way, we will be facing something similar with bash soon. Bash completion v1 (which is what we use for global completion installation) sources entire scripts, but it's deprecated. Bash completion v2 uses a similar lazy-load pattern that does not allow registering a global completion fallback (`complete -D`, which is what we use).

We could always use the alternative strategy of telling the user to source the file directly in their rc script, but that's not a systemwide installation method and it's brittle in other ways.

Testing this branch now.

kislyuk:
Testing looks good, I agree this approach seems to be the best.

kislyuk:
Released in v3.2.0.

PR #471, opened 2024-01-29
Move debug_stream initialization to helper method

### LINKED ISSUE #393: Allow debug_stream to be set optionally
First, a thank you for the project; I've been looking into integrating `argcomplete` into ipython/traitlets#811 so that shell completion works "out-of-the-box" for IPython, Jupyter, and everything else in the ecosystem.

Currently in `CompletionFinder.__init__()`, the following is done:
```python
        global debug_stream
        try:
            debug_stream = os.fdopen(9, "w")
        except Exception:
            debug_stream = sys.stderr
        debug()
```

Could this behavior be made optional? I think it could either be extracted out into a separate helper method that can be overridden, or perhaps be controlled by either an additional argument (similar to `output_stream=`) or environment variable. Feel free to let me know which (or both) is preferable, and I can open a PR.

The problem is that this file descriptor could be used elsewhere, for example by pytest, leading to obscure failures. You can find a MRE in https://github.com/ipython/traitlets/pull/815/commits/2654e3feb9e152148a519c47a73200b9bd683133. I was able to get around this for now by `mock`-ing `os.fdopen` https://github.com/ipython/traitlets/pull/811/commits/0deadccc0adc0d9a91a15fe8790f2b7c5c8831a9, but I'd prefer not to have to introduce mocking if possible.

I think this was also identified as one of two issues in #365 but was closed since the other half was resolved.

azhu-tower:
By default argcomplete tries to use fd 9 as its debug stream, but this could clash with something else trying to use the same file descriptor, for example a pytest plugin.

Split into a _init_debug_stream() helper method which can now be overridden in subclasses.

Closes #393

### DIFF
--- a/argcomplete/finders.py
+++ b/argcomplete/finders.py
@@ -117,11 +117,7 @@ class CompletionFinder(object):
             # not an argument completion invocation
             return
 
-        try:
-            _io.debug_stream = os.fdopen(9, "w")
-        except Exception:
-            _io.debug_stream = sys.stderr
-        debug()
+        self._init_debug_stream()
 
         if output_stream is None:
             filename = os.environ.get("_ARGCOMPLETE_STDOUT_FILENAME")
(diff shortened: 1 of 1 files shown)


kislyuk (changes requested):
Thanks. This looks good, except please do not try to edit Changes.rst manually, it is automatically generated. Can you please remove that change from the PR?

azhu-tower:
Yep updated, got thrown off by the missing bullet point in the most recent entry :)

kislyuk (approved):

kislyuk:
Released in v3.2.3, please test

Issue #255, opened 2018-08-12
test.py relies on python


ignatenkobrain:
Please use sys.executable instead.

evanunderscore:
What problems is this causing for you? Those parts are exercising the global completion code that is matching on `python*` and `pypy*` so I don't think they would work with the full path from `sys.executable`.

evanunderscore:
@hroncok Are you able to comment on whether this has been an issue for you when packaging for Fedora?

hroncok:
Somehow. We have a workaround:

```
sed -i -e "1s\|#!.*python.*\|#!/usr/bin/python3\|" test/prog scripts/*
sed -i -e "s\|python \|python3 \|" test/test.py
```

This makes sure that our packaged python3-argparse RPM is only used for our python3 things.

evanunderscore:
I have a branch which should fix these issues here: https://github.com/evanunderscore/argcomplete/commits/dynamic-test-shebang

It has a few problems:

1. Moving those scripts is rather disruptive to the code base, so I'd want @kislyuk's blessing before doing something like that.
2. The package now has to be installed in order to run the tests since it needs the console_scripts to be created (it could previously read them straight from the `scripts` folder).
3. It's difficult to work this into the CI in a way that we'd know if we broke it again in future (all of the current jobs seem to invoke the interpreter as `python` even when using `python3`).

@hroncok Do you think it's worth pursuing this, or are you happy to keep `sed`ing the code?

hroncok:
I think that it's worth pursuing for multiple reasons\*, however I'm not sure if this is the best way. I haven't yet read the code and will not have the tiem to do it until next week.

Wearing my Fedora, we can live with the current solution. If you only motivation would eb to make our packaging job easier (thanks!), it is probably not worth the effort.

\* I'm not quite sure how argcomplete is _supposed_ to work in "multiple Pythons installed" environment.

evanunderscore:
Until recently, the fact that anything under `scripts/` ran in Python was an implementation detail, and provided that you could run `python` they would work. Unfortunately I didn't notice that #237 now means that `register-python-argcomplete` won't work unless argcomplete is installed in whatever interpreter `python` refers to.

So it seems like we now have two problems:

* Systems don't necessarily have `python` installed any more
* If they do, argcomplete isn't necessarily importable there

I agree these are probably worth fixing. Once you get time to take a look over my branch and confirm the changes work in your environment, we can go from there.

PR #474, opened 2024-02-26
Switch to entry points and pure pyproject.toml config

segevfiner:
See https://github.com/pypa/pipx/issues/1230 & https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#scripts


segevfiner:
@kislyuk

kislyuk:
Sorry, I don't think we need this at this point. Is there a specific problem that you are trying to solve with this? If so, please describe the complete steps to reproduce it.

argcomplete/scripts/activate_global_python_argcomplete.py
```diff
@@ -120,35 +120,40 @@ def link_user_rcfiles():
...
-    print("Defaulting to system-wide installation.", file=sys.stderr)
-    destinations.append(f"{get_zsh_system_dir()}/_python-argcomplete")
-    destinations.append(f"{get_bash_system_dir()}/python-argcomplete")
-
-for destination in destinations:
-    install_to_destination(destination)
-
-if args.dest is None:
-    print("Please restart your shell or source the installed file to activate it.", file=sys.stderr)
+def main():
```
kislyuk, line 123: This breaks the script by moving variables that other functions depend on from global scope into functional scope.

segevfiner:
See the linked issue: https://github.com/pypa/pipx/issues/1230

I'm basically unable to use this on macOS due to that.

kislyuk:
So you are unable to use argcomplete because of a bug in pipx where it's unable to use setuptools scripts when the absolute path to the script contains spaces?

segevfiner:
Yeah. And Homebrew Python started to force `pipx`.

argcomplete/scripts/activate_global_python_argcomplete.py
segevfiner: Fixed it, though there might be cleaner ways to do so.

kislyuk:
I plan to do some testing this weekend and see what I can do about this.

kislyuk:
Sorry, I don't think we will be able to merge this PR.

segevfiner:
@kislyuk Are you sure? Doing this will also allow this to work correctly in Windows...

evanunderscore:
We also have other long-standing issues related to the use of plain scripts that would probably be fixed by this. I'm not up to date with pyproject.toml but a long time ago I proposed moving these to `console_scripts` which I assume is the same thing.

I originally wrote https://github.com/evanunderscore/argcomplete/commits/dynamic-test-shebang to address #255 but the change was never pursued as the person who reported the problem never tested it to verify it addressed their issue. If there is no specific reason to avoid making this change, it would probably be for the better and avoid more issues like this in future.

kislyuk:
@evanunderscore the primary reason this PR was rejected was because it tried to do too many things at the same time, causing me to lose confidence in my ability to review it and in the PR's ability to prove its correctness.

I think migrating to console_scripts is a reasonable move on its own, provided that it's done cleanly and in a staged manner that preserves the ability to review lines of scripts (e.g. first PR transforms the scripts to the format to be used by console_scripts, second PR moves the script file to its desired location and reconfigures setup.py; a separate PR may propose changes to pyproject.toml).

kislyuk:
@segevfiner as described in the documentation (https://github.com/kislyuk/argcomplete?tab=readme-ov-file#support-for-other-shells), windows is not directly supported by argcomplete. Only WSL is supported.

segevfiner:
> @segevfiner as described in the documentation (https://github.com/kislyuk/argcomplete?tab=readme-ov-file#support-for-other-shells), windows is not directly supported by argcomplete. Only WSL is supported.

I'm not sure if any other changes then the `console_scripts` thing is needed though to support it.

I can split up this PR to separate PRs (Or commits, some prefer one PR with separate commits).

kislyuk:
A change to project policy would be needed, and that is outside the scope of this PR.

OK, please split into separate PRs including at least 3 stages as I outlined. I do not want to have one PR incorporating these changes.

PR #475, opened 2024-02-26
Lazy completion loading for Zsh

### LINKED ISSUE #473: Generating lazy completion scripts for Bash/Zsh
It would be nice to be able to generate lazy loaded completion scripts for Bash and Zsh instead of having to eagerly register and execute them at shell startup by `register-python-argcomplete`. For Bash, it is likely the existing script can be used as is, for Zsh, you need `#compdef <command>` as the first line and to not call compdef but rather call the completion function directly, there is probably a trick or two to make the script work as both. I think click-completion uses one https://github.com/click-contrib/click-completion/blob/master/click_completion/zsh.j2, and also cobra https://github.com/spf13/cobra/blob/main/zsh_completions.go, but there are also others, not sure which is the best.

segevfiner:
You can simply save the completion script into a directory in the `$fpath` and it will work, the Homebrew formula for pipx already does that, but it doesn't work without this change

Fixes #473

### DIFF
--- a/argcomplete/shell_integration.py
+++ b/argcomplete/shell_integration.py
@@ -5,7 +5,7 @@
 
 from shlex import quote
 
-bashcode = r"""
+bashcode = r"""#compdef %(executables)s
 # Run something, muting output or redirecting it to the debug stream
 # depending on the value of _ARC_DEBUG.
 # If ARGCOMPLETE_USE_TEMPFILES is set, use tempfiles for IPC.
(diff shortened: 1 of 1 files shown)


segevfiner:
@kislyuk

kislyuk:
Please add some comments next to the conditional describing what zsh_eval_context is expected to be when evaluating a shell completion autoload script vs. a script sourced directly by an rc file or shell command.

segevfiner:
@kislyuk Added a comment

segevfiner:
@kislyuk Added a comment

kislyuk:
Hmm, that is not a comment and it is not what I was looking for.

segevfiner (commented):
Oops, sorry, committed the test code instead of the comment, fixed it now.

kislyuk:
LGTM, thanks.

kislyuk:
Released in v3.2.3, please test

segevfiner:
@kislyuk Seems to work. Though the other problem that I opened https://github.com/kislyuk/argcomplete/pull/474 for is still there of course.

PR #490, opened 2024-06-16
drop python<=3.7 support

kloczek:
According to https://endoflife.date/python python 3.7 has been EOSed 27 Jun 2023.
Filter all code over `pyupgracde --py38-plus`.


docs/conf.py
```diff
@@ -1,42 +0,0 @@
-import os
```
kislyuk, line 1: Why did you delete this file? Please restore it.

kislyuk (changes requested):

kislyuk:
Closing due to lack of response

PR #493, opened 2024-07-07
Convert scripts to the format used by console_scripts

segevfiner:
@kislyuk

Split from #474


segevfiner:
`ruff` seems to have made a backwards incompatible change that broke your CI.

segevfiner:
@kislyuk Did you look at this? Want me to make a separate PR to fix CI?

argcomplete/scripts/activate_global_python_argcomplete.py
```diff
@@ -0,0 +1,165 @@
...
+import argparse
+import os
+import shutil
+import site
+import subprocess
+import sys
+
+import argcomplete
+
+__package__ = "argcomplete"
```
kislyuk, line 20: Why is this set? Please either document a reason why this needs to be set in a comment, or remove this line.

kislyuk (changes requested):
@segevfiner thanks for your contribution, but can you please change the git history such that the move is explicitly reflected? In other words, `git mv` the files from one place to the other, then add new files in their place with no commonality with the original files.

segevfiner:
> @segevfiner thanks for your contribution, but can you please change the git history such that the move is explicitly reflected? In other words, `git mv` the files from one place to the other, then add new files in their place with no commonality with the original files.
> 
> I can't review this PR as it is because there is no diff reflecting any changes (or lack thereof) made in the move.

Just an FYI: A `git mv` is just a `git rm` + `git add` of the new file. There is no special metadata saved in Git about moves, it's always a heuristic. So you might still need to review the individual commits.

argcomplete/scripts/activate_global_python_argcomplete.py
segevfiner: Added a comment

segevfiner:
@kislyuk Fixes committed.

kislyuk:
Thanks, merged

Issue #496, opened 2024-08-02
zsh-related tests fail when TERM variable is set to linux


ogayot:
Hello,
Running the test-suite with TERM=linux causes issues with the zsh-related tests (see below). The TERM variable is set in the Ubuntu CI infrastructure, causing red tests for python-argcomplete.

excerpt:
```
======================================================================
FAIL: test_special_characters (test.test.TestZshGlobalImplicit.test_special_characters)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/olivier/dev/canonical/argcomplete/test/test.py", line 1261, in setUp
    self.assertEqual(output, "ready\r\n")
AssertionError: 'ready\r\n\x1b[m\x0f\x1b[27m\x1b[24m\x1b[J' != 'ready\r\n'
  ready
- 
+ 


======================================================================
FAIL: test_special_characters_double_quoted (test.test.TestZshGlobalImplicit.test_special_characters_double_quoted)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/olivier/dev/canonical/argcomplete/test/test.py", line 1261, in setUp
    self.assertEqual(output, "ready\r\n")
AssertionError: 'ready\r\n\x1b[m\x0f\x1b[27m\x1b[24m\x1b[J' != 'ready\r\n'
  ready
- 
+ 


======================================================================
FAIL: test_special_characters_single_quoted (test.test.TestZshGlobalImplicit.test_special_characters_single_quoted)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/olivier/dev/canonical/argcomplete/test/test.py", line 1261, in setUp
    self.assertEqual(output, "ready\r\n")
AssertionError: 'ready\r\n\x1b[m\x0f\x1b[27m\x1b[24m\x1b[J' != 'ready\r\n'
  ready
- 
+ 


======================================================================
FAIL: test_unquoted_space (test.test.TestZshGlobalImplicit.test_unquoted_space)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/olivier/dev/canonical/argcomplete/test/test.py", line 1261, in setUp
    self.assertEqual(output, "ready\r\n")
AssertionError: 'ready\r\n\x1b[m\x0f\x1b[27m\x1b[24m\x1b[J' != 'ready\r\n'
  ready
- 
+ 


======================================================================
[...]

PR #497, opened 2024-08-02
Ensure ZSH-related tests succeed when TERM variable is set to linux

ogayot:
ZSH tests are failing in the Ubuntu test infrastructure because the TERM variable is set to "linux". Ensure that the test-suite does not fail in this scenario.

I also added the --no-globalrcs option to the zsh invocation ; since --no-rcs only disables loading of $HOME/.zshrc

refs #496


kislyuk:
Thank you!

PR #500, opened 2024-08-10
Fully migrate to pyproject.toml

segevfiner:
Split from #474

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,14 +1,5 @@
 #!/usr/bin/env python
 
-import glob
+from setuptools import setup
 
-from setuptools import find_packages, setup
-
-setup(
-    url="https://github.com/kislyuk/argcomplete",
-    packages=find_packages(exclude=["test", "test.*"]),
-    package_data={"argcomplete": ["bash_completion.d/_python-argcomplete", "py.typed"]},
-    zip_safe=False,
-    include_package_data=True,
-    platforms=["MacOS X", "Posix"],
-)
+setup()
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,14 +1,5 @@
...
 
-import glob
+from setuptools import setup
 
-from setuptools import find_packages, setup
-
-setup(
-    url="https://github.com/kislyuk/argcomplete",
-    packages=find_packages(exclude=["test", "test.*"]),
-    package_data={"argcomplete": ["bash_completion.d/_python-argcomplete", "py.typed"]},
```
kislyuk, line 10: It looks like `package_data` is not migrated

kislyuk (changes requested):

setup.py
segevfiner: You don't need it due to setuptools_scm AFAIK.

kislyuk: That doesn't make any sense. setuptools_scm manages the version, package_data manages including non-Python files in the distribution.

segevfiner: It also includes a hook for `include_package_data`, which is enabled by default in `pyproject.toml` projects, which includes all files tracked by Git unless otherwise excluded.

segevfiner: @kislyuk So would you rather not rely on `setuptools_scm` and would like me to add an explicit `tool.setuptools.package-data` and/or possibly disable `incluede-package-data`?

You can see the docs about this here https://setuptools.pypa.io/en/latest/userguide/datafiles.html#data-files-support

kislyuk:
Thanks for your contribution. I ended up doing equivalent changes as part of migrating the build system from setuptools to hatch (as is now recommended by the Python publishing guide).

Issue #502, opened 2024-09-12
Using environment variables with FilesCompleter


FlorianPommerening:
When completing a filename that includes an environment variable with the `FilesCompleter`, the forwarding to bash doesn't work.

### To reproduce

Test environment:
* Ubuntu 24.04
* Python 3.12.3
* argcomplete 3.1.4-1ubuntu0.1 (this is 3.1.4 with a patch back-ported from 3.3.0 to make it compatible with Python 12.3)
* zsh 5.9

I tested this with the following file `foo.py`

```Python
#!/usr/bin/env python3
# PYTHON_ARGCOMPLETE_OK

import argcomplete, argparse

parser = argparse.ArgumentParser()
arg = parser.add_argument("file")
arg.completer = argcomplete.completers.FilesCompleter()

argcomplete.autocomplete(parser)
```

```Bash
mkdir files
touch files/aa files/ab
export MY_FILES=$(pwd)/files
./foo.py files/<TAB>     # This suggests ['files/aa', 'files/ab'] and completes to 'files/a' as expected.
./foo.py $MY_FILES/<TAB> # This does not offer any suggestions.
```

### Additional Testing in Python

I believe the cause has something to do with the way that `FilesCompleter` forwards the call to bash
https://github.com/kislyuk/argcomplete/blob/develop/argcomplete/completers.py#L67

```Python
 _call(["bash", "-c", "compgen -A file -- '{p}'".format(p=prefix)])
```

I tried to isolate this like this (same setup as above):
```Python
#!/usr/bin/env python3

import subprocess

cmd = ["bash", "-c", "compgen -A file -- '$MY_FILES/'"]
out = subprocess.check_call(cmd, universal_newlines=True)
print(out)
```
Executing this script fails both in zsh and in bash. The test above only failed in zsh, not in bash but I cannot see how it would succeed in bash if calling the `check_call` fails. Maybe I only see the correct completions in bash because the command fails and then falls back to the default completion in bash?

### Additional Testing in zsh/bash

```Bash
bash -c "compgen -A file -- '$MY_FILES/'"
```
Resolves `$MY_FILES` before the call and effectively calls
```Bash
bash -c "compgen -A file -- '/path/to/files/'"
```
and returns
```
/path/to/files/ab
/path/to/files/aa
```
This works both in bash and zsh but is not what the `FilesCompleter` calls and not the completion we expect. We have to escape the `$`.

```Bash
[...]

PR #506, opened 2024-09-27
use interactive shells and bind to work around issue in bash

FlorianPommerening:
This is a workaround to #502 where environment variables in file names are not completed. The issue is that calling `compgen` in a nested bash did not recognize the environment variables. It looks like this is a bug that was fixed in bash 5.3 but it still affects earlier versions (https://savannah.gnu.org/support/index.php?111125) For those earlier version, calling `bind` before the call to `compgen` forces the necessary initialization.

This works with a normal shell, but produces a warning `line editing not enabled` because bind is meant for interactive contexts. Using `-i` to turn the nested shell into an interactive shell avoids the warning. An alternative would be to pipe the warning to `/dev/null` instead.

### DIFF
--- a/argcomplete/completers.py
+++ b/argcomplete/completers.py
@@ -59,13 +59,17 @@ class FilesCompleter(BaseCompleter):
         completion = []
         if self.allowednames:
             if self.directories:
-                files = _call(["bash", "-c", "compgen -A directory -- '{p}'".format(p=prefix)])
+                # Using '-i' and 'bind' in this and the following commands is a workaround to a bug in bash
+                # that was fixed in bash 3.5 but affects older versions. Environment variables are not treated
+                # correctly in older versions and calling bind makes them available. For details, see
+                # https://savannah.gnu.org/support/index.php?111125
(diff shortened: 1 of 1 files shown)


kislyuk:
@FlorianPommerening thanks for digging into this. I like this change and we can call `bind` as advised by upstream. However I'd like to understand better the possible side effects of starting a bash subprocess with the `-i` (interactive) option. The bash manual (https://www.gnu.org/software/bash/manual/bash.html#Interactive-Shells) lists a bunch of things that seem like extras that we don't necessarily want.

If the only purpose of `-i` is to avoid the warning, can we just discard stderr from the shell instead of using `-i`?

kislyuk:
@FlorianPommerening do you want to remove `-i` or should I? The PR otherwise looks ready to merge.

FlorianPommerening:
Whatever you prefer but I won't get around to it this week.

FlorianPommerening:
Sorry this took a while to get around to even if it was a small change but I updated the PR now.

kislyuk:
Thanks for your work on this! Release coming up shortly.

Issue #507, opened 2024-10-05
1 Error and 13 failed tests with Python 3.12.7


mcepl:
With the upgrade to 3.12.7 tests started to fail, most likely because of [large changes to argparse module](https://docs.python.org/3/whatsnew/changelog.html#changelog):

```
[  246s] ======================================================================
[  246s] ERROR: test_repl_parse_after_complete (__main__.TestArgcompleteREPL.test_repl_parse_after_complete)
[  246s] ----------------------------------------------------------------------
[  246s] Traceback (most recent call last):
[  246s]   File "/home/abuild/rpmbuild/BUILD/argcomplete-3.4.0/./test/test.py", line 934, in test_repl_parse_after_complete
[  246s]     args = p.parse_args(["--foo", "spam", "bar"])
[  246s]            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  246s]   File "/usr/lib64/python3.12/argparse.py", line 1895, in parse_args
[  246s]     args, argv = self.parse_known_args(args, namespace)
[  246s]                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  246s]   File "/usr/lib64/python3.12/argparse.py", line 1931, in parse_known_args
[  246s]     namespace, args = self._parse_known_args(args, namespace)
[  246s]                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  246s]   File "/home/abuild/rpmbuild/BUILD/argcomplete-3.4.0/argcomplete/packages/_argparse.py", line 302, in _parse_known_args
[  246s]     start_index = consume_optional(start_index)
[  246s]                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  246s]   File "/home/abuild/rpmbuild/BUILD/argcomplete-3.4.0/argcomplete/packages/_argparse.py", line 168, in consume_optional
[  246s]     action, option_string, _, explicit_arg = option_tuple
[  246s]     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  246s] ValueError: not enough values to unpack (expected 4, got 1)
[  246s] 
[  246s] ======================================================================
[  246s] FAIL: test_choices (__main__.TestArgcomplete.test_choices)
[  246s] ----------------------------------------------------------------------
[  246s] Traceback (most recent call last):
[...]

PR #508, opened 2024-10-06
Fix issue #507

eggry:
In Python 3.13 and 3.12.7, the behavior of `_parse_optional` has changed to return a list of tuples: https://github.com/python/cpython/pull/124631
This change would break the optional argument parsing of this library.

This PR will detect this behavior and return the proper format accordingly.
This PR resolves issue #507.

Cross ref: https://github.com/omni-us/jsonargparse/pull/591#pullrequestreview-2350465411


kislyuk:
Awesome. Thanks for the fix!

PR #512, opened 2024-11-02
Support PYTHON_ARGCOMPLETE_OK_PATHS

### LINKED ISSUE #510: Feature Requets: Environment variable: `PYTHON_ARGCOMPLETE_OK_PATHS`
## Feature Requested

It would be nice if users could specify an environment variable to enable global `argcomplete` on scripts, such as `PYTHON_ARGCOMPLETE_OK_PATHS="/home/me/.local/bin/foo:/home/me/.local/bin/bar"`

This would be equivalent to prepending those two scripts with `# PYTHON_ARGCOMPLETE_OK`

## Motivation

By default, argcomplete works with scripts prefixed with `# PYTHON_ARGCOMPLETE_OK`. This is a good feature, but often inconvenient when the controlling the exact entry point to a program is difficult. For example, `pyproject.toml`'s `project.scripts`

## Alternatives

1. Using `# PYTHON_ARGCOMPLETE_OK`. For scripts installed via `pyproject.toml`, if users use `setuptools` as their backend, they can use `script-files`. As mentioned here (among a few other issues):
https://github.com/kislyuk/argcomplete/issues/371

This solution is lacking in multiple ways, however. First, the documentation explicitly states this should be avoided. Second, when installing in, for example, a virtualenv, setuptools may prefix these scripts with a shebang like `#!/home/me/.virtualenvs/dev/bin/python` which is generate at install-time, not build time when the `project.scripts` must be.

---

2. Using `eval "$(register-python-argcomplete my-python-app)"`. End users can either run `eval "$(register-python-argcomplete my-python-app)"` manually every time, add the command to their bashrc/zshrc files, or if using a virtualenv, perhaps appending that line onto `virtualenv/myvenv/bin/activate`.

Doing so, however, is clunky, puts a good bit of extra expectation on users, and importantly may be quite slow. `eval "$(register-python-argcomplete my-python-app)"` works out to be more than a few lines of code which now need to be executed during every shell login, for every tool; especially compared to 
```bash
PYTHON_ARGCOMPLETE_OK_PATHS="${PYTHON_ARGCOMPLETE_OK_PATHS}:/home/me/.local/bin/foo"
```
---
3. Custom support for setuptools installed scripts. This seems like a viable alternative, but also likely a decent time investment that might end up being difficult given there are multiple issues about supporting pip installed packages. If this is on the TODO list then maybe closing this issue is ok. If it is not, however, I would imagine implementing `PYTHON_ARGCOMPLETE_OK_PATHS` in the meantime would be far simpler. Whichever code scans for `# PYTHON_ARGCOMPLETE_OK` could just check it's environment before scanning.

### Additional Benefits

This is by no means the primary motivation, but a nice benefit of this could be:

This could make it easy for users to turn on or off argcomplete for individual files / paths quickly and easily, if they wanted to, without actually modifying the code of the scripts. This could be extra useful for binaries. For example, I believe `pyinstaller` supports a onefile mode that includes in itself a python binary; adding and removing `# PYTHON_ARGCOMPLETE_OK` from the first 1000 lines of this without re-installing is not something an end-user should ever need to do.

zwimer:
Resolves #510 by supporting `PYTHON_ARGCOMPLETE_OK_PATHS` environment variable

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -40,6 +40,8 @@ follows:
     args = parser.parse_args()
     ...
 
+Alternatively, add the absolute file path to the ``PYTHON_ARGCOMPLETE_OK_PATHS`` environment variable.
+
 Register your Python application with your shell's completion framework by running ``register-python-argcomplete``::
 
     eval "$(register-python-argcomplete my-python-app)"
@@ -210,7 +212,8 @@ Global completion
 -----------------
 In global completion mode, you don't have to register each argcomplete-capable executable separately. Instead, the shell
 will look for the string **PYTHON_ARGCOMPLETE_OK** in the first 1024 bytes of any executable that it's running
-completion for, and if it's found, follow the rest of the argcomplete protocol as described above.
+completion for, and if it's found, follow the rest of the argcomplete protocol as described above. Alternatively, the
+executable path may be added to the environment variable **PYTHON_ARGCOMPLETE_OK_PATHS**.
 
 Additionally, completion is activated for scripts run as ``python <script>`` and ``python -m <module>``. If you're using
 multiple Python versions on the same system, the version being used to run the script must have argcomplete installed.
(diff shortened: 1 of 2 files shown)


zwimer:
@kislyuk If there is anything I can do to help get this merged, or anything else that resolves #510 or the underling issues it defines, please let me know

kislyuk:
Please see discussion in https://github.com/kislyuk/argcomplete/issues/510#issuecomment-2564816215 - I do not expect to merge this PR in its current form.

PR #513, opened 2024-11-18
Preliminary Python 3.13 compatibility

liushuyu:
This pull request adds preliminary Python 3.13 compatibility. A new CI job for Python 3.13 is also added.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04, macos-12, macos-14]
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
 
     steps:
     - uses: actions/checkout@v4
--- a/argcomplete/packages/_argparse.py
+++ b/argcomplete/packages/_argparse.py
@@ -5,6 +5,7 @@
 
 # This file contains argparse introspection utilities used in the course of argcomplete execution.
 
+import sys
 from argparse import (
     ONE_OR_MORE,
     OPTIONAL,
@@ -15,6 +16,7 @@ from argparse import (
     Action,
     ArgumentError,
     ArgumentParser,
+    Namespace,
     _get_action_name,
     _SubParsersAction,
 )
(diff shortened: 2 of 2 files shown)


kislyuk:
Thanks, I think this works!

liushuyu:
> Reverting as this is not actually necessary for tests to pass on Python 3.13.
> 
> Did you have a specific issue with Python 3.13 that this is intended to solve and is not caught by the test suite, or were you talking about Python 3.14?

I am actually one of the Ubuntu maintainers, and our automated testing shows that this project does not work under Python 3.13: https://autopkgtest.ubuntu.com/results/autopkgtest-plucky/plucky/amd64/p/python-argcomplete/20241120_101446_98775@/log.gz. From the issue you linked (#515), it looks like we have the exact same symptoms on Python 3.13 as well.

liushuyu:
From the looks of it, this also fixes #514

paparodeo:
this patch fixes failures of downstream consumers https://cache.nixos.org/log/qzijsl88q4awdsk4w96b91zf9qmvykns-python3.12-traitlets-5.14.3.drv

kislyuk:
@liushuyu I am unable to reproduce the test failures on Python 3.13 that you are referencing. Issue #515 is about Python 3.14, which is not released yet. Can you give me an exact set of commands that I can use to reproduce on Python 3.13?

@paparodeo this package does not support running tests using pytest; please try again while using unittest.

liushuyu:
> @liushuyu I am unable to reproduce the test failures on Python 3.13 that you are referencing. Issue #515 is about Python 3.14, which is not released yet. Can you give me an exact set of commands that I can use to reproduce on Python 3.13?
> 

Because Debian and Ubuntu usually cherry-pick updates from the Python release branch. However, you will be able to reproduce this issue using Python 3.12.8 (released 2 days ago) very soon.

> @paparodeo this package does not support running tests using pytest; please try again while using unittest.

kislyuk:
That's great, thanks. Once I'm able to repro with 3.12.8/3.13.1 I will dig in and figure out what's going on, and will merge a variation of your fix.

liushuyu:
> That's great, thanks. Once I'm able to repro with 3.12.8/3.13.1 I will dig in and figure out what's going on, and will merge a variation of your fix.

I have already did some initial investigations when submitting this pull request, the root cause is this: https://github.com/python/cpython/commit/1fe63b15eb1417503eeba2c638f7296a67eb580c

kislyuk:
@liushuyu thanks for linking that.

argcomplete follows a particular philosophy when patching argparse. The patched class overrides as few functions as possible. Within those functions, it retains as much original code as possible, and clearly marks all code that was modified and the version of CPython standard library the code was copied from.

The task now is for me to verify that the approach you are taking in your patch (overload `_parse_known_args2()`) is in line with that philosophy, and to change it so that it retains as much original code as possible and is annotated.

paparodeo:
> @paparodeo this package does not support running tests using pytest; please try again while using unittest.

the errors were caused by running the tests of the traitlets  module which broke when python was updated to `3.12.8` I believe. https://github.com/ipython/traitlets/tree/main/tests. I don't have any affiliation with that module, just noticed the build started to fail. https://hydra.nixos.org/build/281237249

kislyuk:
Thanks for your work on this @liushuyu. I pursued a slightly different and simpler approach, and released it in argcomplete v3.5.2 which should restore compatibility.

liushuyu:
> Thanks for your work on this @liushuyu. I pursued a slightly different and simpler approach, and released it in argcomplete v3.5.2 which should restore compatibility.

Thank you for the fix!

Issue #514, opened 2024-11-19
Debian patched python 3.12 (or stock python 3.14.0a2) fails testing


yarikoptic:
There is a good number of patches (motivation is unknown to me, didn't track) are in Debian's python 3.12 package. See https://sources.debian.org/patches/python3.12/3.12.7-3/ . Testing of our https://github.com/datalad/datalad/ package started to fail on 3.12 (passed ok before IIRC), and it seems to manifest also in running argcomplete tests:

```shell
❯ ./test/test.py -f
E
======================================================================
ERROR: test_action_activation (__main__.TestArgcomplete.test_action_activation)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/yoh/proj/misc/argcomplete/./test/test.py", line 250, in test_action_activation
    self.assertEqual(set(self.run_completer(make_parser(), cmd)), set(output))
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/misc/argcomplete/./test/test.py", line 139, in run_completer
    completer(parser, output_stream=t, exit_method=sys.exit, **kwargs)
  File "/home/yoh/proj/misc/argcomplete/argcomplete/finders.py", line 174, in __call__
    completions = self._get_completions(comp_words, cword_prefix, cword_prequote, last_wordbreak_pos)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/misc/argcomplete/argcomplete/finders.py", line 223, in _get_completions
    completions = self.collect_completions(active_parsers, parsed_args, cword_prefix)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/misc/argcomplete/argcomplete/finders.py", line 440, in collect_completions
    completions += self._get_option_completions(active_parser, cword_prefix)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yoh/proj/misc/argcomplete/argcomplete/finders.py", line 326, in _get_option_completions
    if not self._action_allowed(action, parser):
[...]

PR #517, opened 2024-12-05
Allow 3.14 as alpha pre-releases already available

yarikoptic:
Ref:
- https://github.com/kislyuk/argcomplete/issues/514#issuecomment-2521133951


kislyuk:
Thanks! Didn't know this was available.

PR #525, opened 2025-03-10
Update shell_integration.py to support spaces in script file path.

wxgeo:
Currently, any space in a script file path will result in an invalid shell code when calling shellcode(..., argcomplete_script=<path containing a space>). Replacing any space with "_SPACE_" solves this issue.


kislyuk:
LGTM, thanks!

PR #526, opened 2025-03-11
Fix completion when wordbreak is first character

ach5948:
When a completion argument starts with a wordbreak character (e.g. "a :b"), argcomplete doesn't properly set `last_wordbreak_pos` and remove the wordbreak character from the results. This causes bash to add an extra wordbreak character (e.g. "a ::b") which then breaks any further completion. This commit properly sets `last_wordbreak_pos` in the _shlex parser, and adds a test case for a leading colon.


ach5948:
Created issue https://github.com/kislyuk/argcomplete/pull/526

kislyuk:
Confirmed this fix works as expected. Thanks!

PR #533, opened 2025-04-13
Improved support statement in README concerning shells and platforms

exploide:
Currently, the argcomplete README gives the reader the impression that the module will exclusively function with bash and zsh. Only if we read nearly towards the end, we learn that it works with other shells, too.

If developers or end-users just skim over the first paragraph, they might think, okay this is not for me, it doesn't work with my favorite shell, and quit. While in fact it would work.

So I propose to make clear early with which shells argcomplete works. I do understand that you want to also make clear that you don't intend to officially support other shells than bash and zsh. So I tried to put emphasis on this.

I hope both points become clear in the README now. Please let me know what you think.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,17 +1,18 @@
-argcomplete - Bash/zsh tab completion for argparse
+argcomplete - shell tab completion for argparse
 ==================================================
 *Tab complete all the things!*
 
 Argcomplete provides easy, extensible command line tab completion of arguments for your Python application.
-
-It makes two assumptions:
-
-* You're using bash or zsh as your shell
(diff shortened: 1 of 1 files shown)


kislyuk:
Thanks for your interest in argcomplete. The opening paragraph of the readme must be clear and space is precious there. I don't think the wording that you propose would work, so I'm going to close this PR now. I do take your point that community support for other shells could be made more discoverable. I'll add a short note to the first bullet point, like "(limited support for other shells exists - see below)".

exploide:
> I'll add a short note to the first bullet point, like "(limited support for other shells exists - see below)".

Thanks. I think this is going to help.

PR #534, opened 2025-04-13
fish: set variable scope to local

exploide:
Small improvement to the fish completion script: Such variables should be explicitly scoped to local. Otherwise the script could accidentally modify a global variable of the same name.

The variables are still exported to the programs invoked by the function so everything works as before but without a potential undesired side-effect.


kislyuk:
Thanks!

Issue #535, opened 2025-04-15
global bash completion breaks completion for some PHP scripts


hroncok:
Hello,

in Fedora, we got a [report from the PHP maintainer](https://bugzilla.redhat.com/show_bug.cgi?id=2359689) that globally installed argcomplete bash completion interferes with completion of some PHP scripts.

To reproduce with pip-installed argcomplete, you need to have some PHP scripts installed. E.g. `pie` on Fedora 41:

```
$ podman run --rm -ti fedora:41 /usr/bin/bash
...
# dnf -y install bash-completion pip pie
...
# bash          # this is needed to make bash-completion of pie work
# pie lis[tab]  # completes to: pie list

# pip install argcomplete
...
Successfully installed argcomplete-3.6.2
...
# activate-global-python-argcomplete
...
# exit
# bash
# pie lis[tab]  # completes to nothing
```

remicollet:
Checking if interpreter is python is a possible fix (but I don't know a lot about other possible python interpreters, and I'm know a bash_completion expert)

```
            elif "${interpreter[@]}" --version 2>/dev/null | grep -qi python \
                && __python_argcomplete_run "${interpreter[@]}" -m argcomplete._check_console_script "$SCRIPT_NAME"; then
```

hroncok:
A similar check already exists in https://github.com/kislyuk/argcomplete/blob/ab03e8a8dd9313083728eb980bb621df1c940f2c/argcomplete/bash_completion.d/_python-argcomplete#L154 but this is in an elif branch of that check. Adding that grep would effectively mean that it never happens.

hroncok:
I don't really understand what is the purpose of the elif branch -- scripts with non-Python shebangs using argcomplete?

hroncok:
A check for the interpreter's name seems to work.

Not sure if running the interper with --version and prepping it for python (as suggested by @remicollet) is better or worse.

```patch
--- a/argcomplete/bash_completion.d/_python-argcomplete
+++ b/argcomplete/bash_completion.d/_python-argcomplete
@@ -193,7 +193,8 @@ _python_argcomplete_global() {
             if (__python_argcomplete_scan_head_noerr "$SCRIPT_NAME" easy_install \
                 && "${interpreter[@]}" "$(__python_argcomplete_which python-argcomplete-check-easy-install-script)" "$SCRIPT_NAME") >/dev/null 2>&1; then
                 ARGCOMPLETE=1
[...]

PR #536, opened 2025-04-17
Only execute Python interpreters

hroncok:
Fixes https://github.com/kislyuk/argcomplete/issues/535

This checks if the interpreter is anything with python or pypy in it.

Thus, avoiding executing e.g. `php -m argcomplete._check_console_script ...` (which succeeds).


hroncok:
I can attempt to write a test for this, however, not today.

kislyuk:
Thank you!

hroncok:
> I can attempt to write a test for this, however, not today.

I attempted to create a script with a shebang that passes even with the `-m` flag and succeeds.

```
#!/usr/bin/env echo
```

Such script can then be used like this:

```
    def test_foreign_scripts_not_broken(self):
        self.sh.run_command("cd " + TEST_DIR)
        self.sh.run_command("complete -W 'hello' echoer")
        self.sh.run_command(TestBashZshGlobalBase.install_cmd)
        self.assertTrue(self.sh.run_command("echoer h\t").endswith("echoer hello\r\n"))
```

However, I am not able to make it fail without the fix, so it is not useful as a test :/

PR #543, opened 2025-08-12
Update fish shell documentation

samdoran:
Completions using a path must have the full path and be sourced in order to work correctly.


kislyuk:
Thanks!

PR #551, opened 2026-05-30
quote prefix passed to compgen in FilesCompleter

sahvx655-wq:
FilesCompleter interpolates the prefix into a single-quoted compgen string, so a quote in the completed word breaks out and runs arbitrary shell.


sahvx655-wq:
any update?

kislyuk:
Thanks. This needs a test, but I can make one myself

PR #552, opened 2026-06-03
escape glob and brace metacharacters in quote_completions

sahvx655-wq:
noticed while reading quote_completions that the unquoted branch escapes \* but not the other expansion metacharacters. bash treats them as the same class:

    $ printf '%q\n' 'data?' 'a{b,c}'
    data\?
    a\{b\,c\}

so a completed filename like data? or {x,--flag} stays literal on the line but gets glob/brace expanded by the shell on execution, changing which files or arguments the command receives. escape ? [ ] { } alongside \* in the unquoted case so completions go in as literal tokens.


kislyuk:
Thank you!

PR #553, opened 2026-07-09
Expand ~ in DirectoriesCompleter (fixes #529)

### LINKED ISSUE #529: DirectoriesCompleter not work with ~
```py
parser = argparse.ArgumentParser(prog='test, add_help=False)
parser.add_argument('directory', type=dir_path).completer = DirectoriesCompleter()

autocomplete(parser)
args = parser.parse_args()
```

`test ~/<TAB>`
show files in home directory

(`os.listdir("~")` raise `FileNotFoundError: [Errno 2] No such file or directory: '~'`)

apoorvdarshan:
## Summary

`DirectoriesCompleter` (and its base `_FilteredFilesCompleter`) yields no
completions for a tilde-prefixed path such as `~/`, so `~/<TAB>` shows nothing
even though the home directory has subdirectories. Reported in #529.

The cause is in `_FilteredFilesCompleter.__call__`: `os.path.dirname("~/")`
returns `"~"`, and `os.listdir("~")` raises `FileNotFoundError` (there is no
literal `~` directory). That exception is swallowed by the surrounding
`except Exception: return`, so the completer silently returns an empty iterator.

## Fix

Expand the prefix with `os.path.expanduser` before splitting and listing, so
the filesystem operations run against the real home directory. This mirrors the
behavior of the bash-backed `FilesCompleter`, whose `compgen` already expands
`~`. Non-tilde prefixes are unaffected (`expanduser` is a no-op for them).

```python
expanded_prefix = os.path.expanduser(prefix)
target_dir = os.path.dirname(expanded_prefix)
...
incomplete_part = os.path.basename(expanded_prefix)
```

## Tests

Added `test_directory_completion_with_tilde` (regression test for #529). It
points `HOME`/`USERPROFILE` at a temp directory so the test is hermetic, then
asserts that `~/`, `~/ab`, and `~/abc/` complete correctly. The test fails on
the unpatched code (empty result) and passes with the fix.

- `python -m unittest test.test.TestArgcomplete` — 36 passed
- `ruff check argcomplete test/test.py` — clean
- `ruff format --check argcomplete/completers.py test/test.py` — clean
- `mypy --install-types --non-interactive argcomplete` — clean

Disclosure: prepared with AI assistance; reviewed and verified locally.

### DIFF
--- a/argcomplete/completers.py
+++ b/argcomplete/completers.py
@@ -104,12 +104,15 @@ class _FilteredFilesCompleter(BaseCompleter):
         """
         Provide completions on prefix
         """
-        target_dir = os.path.dirname(prefix)
+        # Expand a leading "~" so that filesystem operations below work on a
+        # real path (os.listdir("~") would otherwise raise FileNotFoundError).
+        expanded_prefix = os.path.expanduser(prefix)
+        target_dir = os.path.dirname(expanded_prefix)
         try:
             names = os.listdir(target_dir or ".")
         except Exception:
             return  # empty iterator
-        incomplete_part = os.path.basename(prefix)
+        incomplete_part = os.path.basename(expanded_prefix)
         # Iterate on target_dir entries and filter on given predicate
         for name in names:
             if not name.startswith(incomplete_part):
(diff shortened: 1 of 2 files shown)


kislyuk:
Thanks!

PR #554, opened 2026-07-15
100% type coverage

jorenham:
Hi there.

I noticed that `argcomplete` was missing some type annotations, with a type coverage of 26.6% (according to `uvx pyrefly coverage check argcomplete --public-only` ([docs](https://pyrefly.org/en/docs/report/))).

I figured I might as will contribute and bring this too 100%, so here I am :). 

Mypy still passes with flying colors, and `pyrefly coverage check` now reports a shiny 100% type coverage:

```console
$ mypy argcomplete
Success: no issues found in 15 source files

$ uvx pyrefly coverage check argcomplete --public-only
 INFO type coverage 100.00% (94 of 94 typable)
```

I realize this might be a bit much for a single PR, so if you want, I could split this up.

Oh and in case you were wondering; I'm a human, and I did not use any AI for this; it's way too much fun to do myself :)

### DIFF
--- a/argcomplete/__init__.py
+++ b/argcomplete/__init__.py
@@ -9,5 +9,5 @@ from .io import debug, mute_stderr, warn
 from .lexers import split_line
 from .shell_integration import shellcode
 
-autocomplete = CompletionFinder()
+autocomplete: CompletionFinder = CompletionFinder()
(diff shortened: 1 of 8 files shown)


kislyuk:
Thanks!

PR #555, opened 2026-07-20
Enforce 100% type coverage in CI

jorenham:
Now that `argcomplete` has 100% public type coverage (#554), I figured it might be a good idea to ensure that it stays that way.

This adds a light-weight job to the CI workflow that runs `pyrefly coverage check`, which will exit 1 if the type coverage dips below 100%.

If that ever happens, e.g. because of an API change, or because Pyrefly changed something (which I don't expect will happen), then feel free to ping me :)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,3 +43,10 @@ jobs:
       - uses: astral-sh/ruff-action@v4.0.0
         with:
           args: "format --check"
+  pyrefly-coverage:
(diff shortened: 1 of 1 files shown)


kislyuk:
sorry we will not be adding a dependency on pyrefly to this project.

jorenham:
> sorry we will not be adding a dependency on pyrefly to this project.

may I ask way?

PR #558, opened 2026-08-04
Require Python 3.10+

### LINKED ISSUE #559: Version incompatibility bug breaking CI
https://github.com/kislyuk/argcomplete/blob/da6d0552fcf0f6218e4431b564f8c9d3f5a543d4/argcomplete/completers.py#L35

This is Python 3.10+ functionality but the project dependencies say Python 3.8+ (Solved by #558). I would merge that PR and yank the previous release.'

Also this project should add some linting to prevent this. `ruff` does this kind of checking already and is miles faster than `black` and `isort` which are currently used.

austin3dickey:
Hi! 👋 This PR raises `requires-python` from `>=3.8` to `>=3.10` and drops the stale classifier. Fixes https://github.com/kislyuk/argcomplete/issues/559.

I noticed with the latest release that `argcomplete` uses unions in type annotations, so importing it raises `TypeError` on Python 3.9 and older. But the published metadata still allowed python 3.8, so resolvers install version 3.7.1 into python 3.9 environments, and any import of `argcomplete` fails there.

Downstream projects that still support 3.9 for some reason would currently need a version exclusion to keep 3.9 jobs working, but a patch release after this PR (and yanking 3.7.1) would let them drop it. Thanks!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -253,7 +253,7 @@ work for zsh as well.
 
 Python Support
 --------------
-Argcomplete requires Python 3.9+.
+Argcomplete requires Python 3.10+.
 
 Support for other shells
 ------------------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 name = "argcomplete"
 description = "Bash tab completion for argparse"
 readme = "README.rst"
-requires-python = ">=3.8"
+requires-python = ">=3.10"
 license = { text = "Apache Software License" }
 authors = [{ name = "Andrey Kislyuk"}, {email = "kislyuk@gmail.com" }]
 maintainers = [{ name = "Andrey Kislyuk"}, {email = "kislyuk@gmail.com" }]
@@ -15,7 +15,6 @@ classifiers = [
   "Operating System :: POSIX",
   "Programming Language :: Python",
   "Programming Language :: Python :: 3",
-  "Programming Language :: Python :: 3.9",
   "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",


zoumingzhe:
Annotations can be modified and compatible with Python 3.8. https://github.com/kislyuk/argcomplete/pull/560

We should not easily upgrade Python version require, as there are still many production environments using Python 3.9.

austin3dickey:
That works too. My understanding was that the maintainers didn't plan to support Python 3.8 or 3.9 because @kislyuk dropped them from CI two weeks ago (in https://github.com/kislyuk/argcomplete/commit/0d1cb9ad7284967f34a02fbaf131525b66f819b0). But I see you're adding them back in #560.

henryiii:
FYI, 3.8 and 3.9 now break, which is generally not what "dropping them" means. If you make a release requiring 3.10+, you can then 'yank' the broken on 3.8 and 3.9 release, which will fix the 3.8/3.9 users and 3.10 users will get the new version. (This PR, but it does require yanking to work). Another option is to restore 3.8 (#560), release, then drop 3.8/3.9 while updating the requires-python setting at the same time.

henryiii:
(argcomplete is pulled in by nox, and nox has dropped Python 3.8 / Python 3.9, so it's quite tricky for use to fix those from nox!)

austin3dickey:
Agreed; I edited the PR description to include yanking 3.7.1.

henryiii:
> We should not easily upgrade Python version require, as there are still many production environments using Python 3.9.

BTW, dropping 3.9 (once this is fixed) would keep this from happening in the future; these systems are legacy systems and don't need the latest features, they need stability. If everyone dropped old Pythons at exactly the same time, that would reduce the breakage for old systems. Yes, no new features, but you shouldn't be running an EoL Python and wanting the latest features at the same time. ;)

kislyuk:
Thanks.

kislyuk:
Thank you all for the discussion; I will not be yanking any releases, but I will roll a patch release shortly!

henryiii:
> but I will roll a patch release shortly!

Reminder, that if you don't release a patch release that restores Python 3.8/3.9 temporarily (or yank), that won't fix anything and all legacy code using 3.8/3.9 will continue to be broken.

Yanking does not stop anyone from explicitly downloading the version, it just removes it from open resolves as a "best version", and this is exactly what it was designed for.

ktbarrett:
@kislyuk the alternative to yanking is everyone has to forever add `argcomplete!=7.3.1` to their deps.

kislyuk:
I released v3.7.2 and yanked v3.7.1 as an exception.

Everyone does not have to forever add anything. Python versions that are past their EOL date are unsupported.

henryiii:
Really appreciate it! It's okay to stop releasing features and even fixes for old versions! It's just nice to "leave working alone" so old versions continue to work (which is what you've done now, thanks again!!!). I still have places like pybind11 & scikit-build-core that haven't dropped 3.9 yet (following manylinux rather than EoL, due to supporting compilation) and use nox, and it's great that I don't have to manually patch everything.

> forever add

Agree it's not forever. :)

PR #560, opened 2026-08-05
update type hints for better Python compatibility

zoumingzhe:
class ChoicesCompleter(BaseCompleter):
>       choices: Final[Mapping[str, str | bytes]]
E       TypeError: unsupported operand type(s) for |: 'type' and 'type'

/root/.pyenv/versions/3.9.18/lib/python3.9/site-packages/argcomplete/completers.py:35: TypeError


kislyuk:
Thank you for your PR. I appreciate the effort; unfortunately I won't be merging it in its current form because we will not be supporting EOL Python versions.

Issue #563, opened 2026-08-19
short-options don't parse correctly without a space


rokke-git:
a number of bugs related to short-options.

mwe.py:
```python
#!/usr/bin/env python3
from argparse import ArgumentParser
from argcomplete import autocomplete

parser = ArgumentParser()
parser.add_argument(
    '--option',
    '-o',
    choices=['one', 'two', 'three'],
)
parser.add_argument(
    '--bool',
    '-b',
    action='store_true',
)
autocomplete(parser)
options = parser.parse_args()

print(options.option)
```

complaints:
```
$ ./mwe.py -o'o<tab>
$ ./mwe.py -o'o       # nothing completes. should return `-o'one' `

$ ./mwe.py -o't<tab>
$ ./mwe.py -o't       # nothing completes. should show the options `two` and `three`

$ ./mwe.py -o'<tab>
$ ./mwe.py -o'-o'     # huh?

$ ./mwe.py -bo<tab>   # nothing completes. should have added a space, or shown `one` `two` `three`
$ ./mwe.py -boo<tab>  # nothing completes. should return `-boone `
```

PR #564, opened 2026-09-12
Complete values attached to unquoted short options

codeofwxz:
Related #563.

Unquoted attached values such as `-oo` and `-boo` currently return no completions for `-o` with choices `one`, `two`, and `three`. This change resolves the current word using argparse's option lookup, walks valid leading switches, and completes the value with its original option prefix (`-oone` or `-boone`). It also handles a cluster ending in a value-taking option, such as `-bo`.

Exact option names and ambiguous abbreviations retain argparse's precedence. The lookup respects nargs and mutual exclusion; leading switches contribute to the completer's local namespace only when their action class is already in `safe_actions`. Custom value completers receive the value prefix, and completion descriptions retain the reconstructed option prefix.

This is a partial fix: completion inside an unclosed quote, including the quoted examples in #563, remains on the existing path. Shell replacement boundaries for those examples are not addressed here.

Initial revision validation on Windows with Python 3.12.14 and 3.14.4:

- Nine new unittest methods covering 28 cases pass, including choices, clusters, nargs, option precedence, custom completers, unsafe actions, subparsers, mutual exclusion, and negative attached values. Fourteen cases fail against the unchanged production baseline.
- Ruff check, Ruff format check, mypy, and wheel build pass.
- The full upstream unittest suite was attempted on both Python versions. The baseline has 191 tests and this branch 200; both report the same 2 failures, 131 errors, 12 skips, and 1 expected failure. The failing/error test identities and exception categories match. These are not full-suite passes.
- A local wrapper resolves only the test module's import-time Bash version probe to existing Git Bash, since bare `bash` selects an unavailable Windows launcher. The wrapper is not part of this patch. Real interactive Bash/Zsh/pexpect Tab insertion and the Linux/macOS CI matrix have not been validated locally.

Review revision: attached values now also complete when ExclusiveCompletionFinder has already seen the option in an earlier word. The change retains custom finder constraints, mutual exclusion, and the current cluster's count context. Thirteen targeted unittest methods pass on Python 3.14, with Ruff lint/format checks passing. This revision was not rerun through the full suite or an interactive shell Tab test; the initial validation and limitations above remain historical results for the initial revision.


codeofwxz:
Thanks for calling out the ExclusiveCompletionFinder interaction. The first-use `prog -oo` case already completes successfully: the current word is excluded from the words passed to parse_known_args, so that action has not been recorded as seen yet.

There was a related gap when an option had appeared in an earlier word: `prog -o one -oo` and `prog -v -bvvoo` returned no attached completion with ExclusiveCompletionFinder, although their separated-value equivalents worked. This revision fixes those cases by relaxing only the seen-option suggestion filter while checking an explicitly typed attached value. It still dispatches through the finder override and preserves mutual exclusion. The new regressions fail on those two cases before the fix and pass afterward.

The count replay should remain: switches in the current word have not been parsed. Protocol tests confirm `prog -bvvoo` supplies v=2, and `prog -v -bvvoo` supplies v=3. Removing that replay would lose the current cluster's count and boolean context.

Validation: 13 targeted unittest methods passed on Python 3.14, plus Ruff lint/format checks. This revision was tested through the completion protocol, without a full suite run or interactive shell Tab test.

